Issue #11, opened 2022-03-30
Highlight changes in changed API items more clearly


Enselic:
Instead of  this for example:

* `pub fn syntect::parsing::Scope::is_empty(&self) -> bool` became <br/> `pub fn syntect::parsing::Scope::is_empty(self) -> bool`

we should somehow highlight more clearly that the only difference is the removal of a `&`. Maybe something similar to what https://docs.rs/pretty_assertions/latest/pretty_assertions/ does, using colors for example.

This request came from @douweschulte in #4.

Issue #4, opened 2022-02-15
Syntax highlighted output


Enselic:
It would be nice to syntax highlight the output. Should be pretty easy.
1. Create a custom sublime syntax for the output format
2. Use the bat library to highlight the output

douweschulte:
![image](https://user-images.githubusercontent.com/25853274/160824677-b935f3ff-21cb-48e5-a0a2-47d863c52b30.png)
I do agree it looks nicer that way. Here I just bodged my way until it worked. I used the `syntect` crate to supply the highlighting rules. If you want to see how I did it, see my fork: <https://github.com/douweschulte/public_items>.

I would like to add that for me highlighted differences in the changes list would be really nice, maybe even more than syntax highlighting. Something like the removed line the changed text in red and on the added line the text in green.

Enselic:
Super big thanks for prototyping that! That indeed looks much better. I looked at how you implemented it, and have the following comments. I understand you just did a quick and dirty proof of concept, but I would like to give some feedback anyway :)
* I would prefer to have the highlighting performed in the convenient `cargo public-items` wrapper (i.e. the project this issue is in) rather in the core `public_items` library, that I would like to be focused on raw processing of public items.
* For robustness I think we need to write a custom syntax definition rather than using the real Rust syntax.
* The exact textural representation of public items is still subject to change and tweaks. In particular, fixing https://github.com/Enselic/public_items/issues/16 will probably change it a bit. But some aspects are of course extremely unlike to change. 

I agree a more user-friendly changed-item diff would be great to have. I filed https://github.com/Enselic/cargo-public-items/issues/11 for it.

douweschulte:
Thanks for the feedback points. But how would a custom syntax be more robust than the real Rust syntax? I would think that catching all edge cases etc would be pretty hard. I would guess that using some library would be less work, especially in the light of syntax changes/additions in upcoming versions of Rust.

I would agree that pure highlighting could just as well or beter live in the wrapper. But maybe it would be nice for other uses of the `PublicItems` to know its composition. I was thinking about something like `syn`, but maybe that is a bit OP. But having the public items fully tokenized could give rise to extra uses. Like having the comparison for Enselic/public-api#11 done on a token level instead of text level, which would make it more robust. This could also potentially make it possible to find out if a change is really breaking (the function accepts less things) or something else. But that is maybe not your vision for this project.

I would be happy to help. I am quite interested in where this project could go.

Enselic:
[...]

PR #12, opened 2022-03-31
Syntax highlighting (#4 & #11)

douweschulte:
This is my current state of the implementation of #4 and #11.
image
Merge needs to wait until these PRs are merged in Enselic/public_items:
* Enselic/public_items#53 with the support for tokenization.
* Enselic/public_items#63 with the support for token based diffing.

### DIFF
--- a/src/markdown.rs
+++ b/src/markdown.rs
@@ -52,188 +52,28 @@ fn print_items_with_header<W: std::io::Write, T>(
     writeln!(w)
 }
 
-use ansi_term::{ANSIString, ANSIStrings, Colour, Style};
+use ansi_term::{ANSIString, ANSIStrings, Colour};
 use public_items::tokens::*;
 
 fn print_with_colour(item: &public_items::PublicItem) -> String {
-    if let Ok(tokens) = item.tokens() {
-        let mut output = Vec::new();
-        let space = Style::new().paint(" ");
-
-        for qual in &tokens.qualifiers {
-            output.push(Colour::Blue.paint(qual.to_string()));
-            output.push(space.clone());
-        }
-
-        let mut path = Vec::new();
-        for id in &tokens.path {
-            path.push(Colour::Blue.paint(id));
-            path.push(Style::new().paint("::"));
-        }
-        path.remove(path.len() - 1);
-
-        match &tokens.kind {
-            Kind::Function {
-                generics,
-                arguments,
-                return_type,
-            } => {
-                output.push(Style::new().paint("fn "));
-                output.extend(path);
-                if !generics.params.is_empty() {
-                    output.push(Style::new().paint("<"));
-                    for param in &generics.params {
-                        output.push(Colour::Blue.paint(param.name.clone()));
-                        output.push(Style::new().paint(": "));
-                        output.push(Colour::Red.paint("generic stuff")); // TODO: add better support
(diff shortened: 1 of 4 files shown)


src/markdown.rs
```diff
@@ -29,7 +29,7 @@ pub fn print_diff(w: &mut impl std::io::Write, diff: &PublicItemsDiff) -> std::i
         w,
         "## Added items to the public API",
         &diff.added,
-        |w, item| writeln!(w, "* `{}`", item),
+        |w, item| writeln!(w, "* {}", print_with_colour(item)),
```
Enselic: Maybe we should output regular (colored) text by default and make markdown opt-in? I'm not sure it makes sense to put colors in markdown. Let me know if you want me to write the code to select output format so you can focus on tokenization and coloring.

douweschulte: That sounds lovely. If you have an idea how to tackle this, that would be very nice. Otherwise I will work on that after #11.

Enselic: Here is the PR (I pinged you in it): https://github.com/Enselic/cargo-public-items/pull/14

douweschulte:
Sneek peak:
syntax_highlighting

I am quite happy with this. I need to test how it handles other inputs so that needs some time. After that I will go and clean up the stuff that is not needed. I envision to remove the prefix/path/suffix from the `PublicItem` if possible en replace it with the fully tokenized output, let me know what you think about that.

Enselic:
Nice! Colors makes such a big difference. Looking forward to the next update.

> I envision to remove the prefix/path/suffix from the PublicItem if possible en replace it with the fully tokenized output, let me know what you think about that.

That sounds perfectly reasonable on a conceptual level, so please go ahead with that 👍 I'm happy to answer any questions that might arise from such an operation.

douweschulte:
image

I made a start on #11 see above. The code is not really clean yet. I am not really sure how to style it in a way that works best for other people, so if you have ideas that would be welcome.

Enselic:
Yeah it's not obvious how to present diffs in a good way. Maybe it would make sense to get regular coloring in place first, and then after that we can iterate on how to show diffs?

In your example above, I think one improvement would be to have the green color on the + line, and red color on the - line. So the added `impl` should have green background, and the removed `String,` should have red background, rather than having the color on the "opposite" side.

Another source to draw inspiration from could be `git show/diff --color-words`, but I'm not sure I like that style:

Skärmavbild 2022-04-02 kl  09 36 58

douweschulte:
image
Okay this looks way better, the background colour does not work that well because it is way to bright and I do not see a way to get dimmed background colours. So I am thinking about doing other style things, like strikethrough or italic. But this means that the `tokens::ChangedTokenStream` is pretty much where I want it.

Enselic:
Hi again! Just an update. I have now
* Renamed public_items to public-api
* Released public-api v0.9.0 that contains your new tokens API
* Released cargo-public-api v0.9.0 that depends on the new public-api lib, but does not make use of tokens yet.

It would be fantastic if you could create a new PR (I think that will be more clean than updating this one) that gets syntax highlighting in place. I would really like us to do that separately from the intra-item diffing, so keep things focused and incremental. Again, no rush, feel free to take your time with that :)

douweschulte:
Okay sounds like a plan. I will get on it when I have time.

PR #14, opened 2022-04-01
Support output formats "plain" (with and without colors) and "markdown"

Enselic:
This PR makes diffing use colors by default:
Skärmavbild 2022-04-01 kl  17 09 04

But if you pipe to a file (or pass `--color=never`) then the tool will use + and - instead of red and green:
Skärmavbild 2022-04-01 kl  17 09 22

### DIFF
--- a/src/arg_types.rs
+++ b/src/arg_types.rs
@@ -1,0 +1,67 @@
+use anyhow::anyhow;
+
+use std::str::FromStr;
+
+use crate::{markdown::Markdown, output_formatter::OutputFormatter, plain::Plain};
+
+#[derive(Debug)]
+pub enum OutputFormat {
+    Plain,
+    Markdown,
+}
+
+impl FromStr for OutputFormat {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s == "plain" {
+            Ok(OutputFormat::Plain)
+        } else if s == "markdown" {
+            Ok(OutputFormat::Markdown)
+        } else {
+            Err(anyhow!("See --help"))
+        }
+    }
+}
+
+impl OutputFormat {
+    pub fn formatter(&self) -> Box<dyn OutputFormatter> {
+        match self {
+            OutputFormat::Plain => Box::new(Plain),
+            OutputFormat::Markdown => Box::new(Markdown),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum Color {
+    Auto,
+    Never,
+    Always,
+}
+
+impl FromStr for Color {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s == "auto" {
+            Ok(Color::Auto)
+        } else if s == "never" {
+            Ok(Color::Never)
+        } else if s == "always" {
+            Ok(Color::Always)
+        } else {
+            Err(anyhow!("See --help"))
+        }
+    }
+}
+
+impl Color {
+    pub fn active(&self) -> bool {
+        match self {
+            Color::Auto => atty::is(atty::Stream::Stdout), // We should not assume Stdout here, but good enough for now
+            Color::Never => false,
+            Color::Always => true,
+        }
+    }
+}
--- a/src/arg_types.rs
+++ b/src/arg_types.rs
@@ -14,12 +14,10 @@ impl FromStr for OutputFormat {
     type Err = anyhow::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        if s == "plain" {
-            Ok(OutputFormat::Plain)
-        } else if s == "markdown" {
-            Ok(OutputFormat::Markdown)
-        } else {
-            Err(anyhow!("See --help"))
+        match s {
+            "plain" => Ok(OutputFormat::Plain),
+            "markdown" => Ok(OutputFormat::Markdown),
+            _ => Err(anyhow!("See --help")),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


Enselic:
@douweschulte Would you like to code review this? You absolutely don't have to of course, but I wanted to give you a chance. I would love to get your feedback on it, even if you don't want to do a formal review, which is perfectly fine. Again, no obligations, we're just having fun :)

The current usage of colors is just a proof of concept. You are free to change colors however you see fit as part of the work you currently do.

If you think it looks reasonable, then I will make tests more robust. (I plan on doing https://github.com/Enselic/public_items/pull/46 also for `cargo-public-items` first in that case.)

src/arg_types.rs
```diff
@@ -0,0 +1,67 @@
...
+    Auto,
+    Never,
+    Always,
+}
+
+impl FromStr for Color {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s == "auto" {
```
douweschulte: Couldn't this be refactored to a match construct?
```rust
match s {
    "auto" => Ok(Color::Auto),
    "never" => Ok(Color::Never),
    "always" => Ok(Color::Always),
    _ => Err(anyhow!("See --help")),
}
```

src/arg_types.rs
```diff
@@ -0,0 +1,67 @@
...
+pub enum OutputFormat {
+    Plain,
+    Markdown,
+}
+
+impl FromStr for OutputFormat {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s == "plain" {
```
douweschulte: Couldn't this be refactored to a match construct?
```rust
match s {
    "plain" => Ok(OutputFormat::Plain),
    "markdown" => Ok(OutputFormat::Markdown),
    _ => Err(anyhow!("See --help")),
}
```

src/markdown.rs
```diff
@@ -1,53 +1,48 @@
...
-
-    Ok(())
-}
+use crate::{output_formatter::print_items_with_header, Args, OutputFormatter};
+
+pub struct Markdown;
+
+impl OutputFormatter for Markdown {
+    fn print_items(&self, _w: &mut dyn Write, _args: &Args, _items: Vec<PublicItem>) -> Result<()> {
+        todo!("Not yet implemented because I'm not sure what the output should look like. Feel free to open a PR with a proposal.")
```
douweschulte: Would a plain list just work? At least for now. Grouping related methods/structs (all struct fields as a bullet list?) would be really cool, but that should just work the same as the diff variant.

I would totally agree that this is not a very useful feature, I would value `docs.rs` above a plain list. But creating a plain list is maybe better than a panic?
```rust
print_items_with_header(
    w,
    "## Public API",
    &items,
    |w, item| writeln!(w, "* `{}`", item),
)?;
```

src/output_formatter.rs
```diff
@@ -0,0 +1,27 @@
...
+use public_items::{diff::PublicItemsDiff, PublicItem};
+
+use crate::Args;
+
+pub trait OutputFormatter {
+    fn print_items(&self, w: &mut dyn Write, args: &Args, items: Vec<PublicItem>) -> Result<()>;
+    fn print_diff(&self, w: &mut dyn Write, args: &Args, diff: &PublicItemsDiff) -> Result<()>;
+}
+
+pub fn print_items_with_header<T>(
```
douweschulte, line 12: Would this fit better as a part of the above trait? Just as a [default implemented method](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations). If you provide the implementation in the trait itself the implementers need not be changed.

douweschulte (approved):
Looks like a nice reorganisation! I left some comments in the files.

src/arg_types.rs
Enselic: Thanks, good catch, fix pushed

src/arg_types.rs
Enselic: Thanks, good catch, fix pushed

src/markdown.rs
Enselic: Yep I had some kind of plain list in mind, but wasn't sure which variant to use. Now that you have cast your vote, I agree it makes good sense, so I took your code and put it in a commit with you as the author (it felt wrong to put me as author).

Grouping should indeed be handled universally by `public_items` rather than by `cargo public-items`, i.e. not only when diffing, but also when just listing. The issue for it is https://github.com/Enselic/public_items/issues/34. I've been trying to figure out for a while how to solve that in a good way, but it's tricky. I'm pretty sure your tokenization of `PublicItem` will make it easier to solve that.

I actually find plain lists of public items quite useful, but then I would use "plain" output format and not "markdown". I've found myself doing things like `cargo public-items | grep public_items::diff` for example.

src/output_formatter.rs
Enselic: Interesting idea, I tried it out, but got:

```
7  | pub trait OutputFormatter {
   |           --------------- this trait cannot be made into an object...
...
11 |     fn print_items_with_header<T>(
   |        ^^^^^^^^^^^^^^^^^^^^^^^ ...because method `print_items_with_header` has generic type parameters
```

and I think removing the generic type parameter is too big of an operation at the moment. So if you don't mind I think we can keep it as it is.

Enselic:
Big thanks for code review! I addressed all comments now. As a next step I will add tests for this change before I merge it, which will require me to land some test-refactorings first, mainly turning `test-invocation-variants.sh` into proper `cargo test` tests.

src/output_formatter.rs
douweschulte: I overlooked that then, I think it is fine where it is.

douweschulte:
Nice! Looking forward to it!

Enselic:
Ok, tests added. I feel ready to merge, but I wanted you to give a chance to review my tests before merge. Doing another review round is completely optional for you of course.

douweschulte (approved):

PR #23, opened 2022-04-16
Perform syntax highlighting on output format `plain`

douweschulte:
I moved all changes, related to Tokens, from the the other PR (#12) into a single commit. And for consistency I renamed all `Colour` to `Color`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,5 @@
+#![deny(clippy::all, clippy::pedantic)]
+
 use std::io::stdout;
 use std::path::{Path, PathBuf};
 
--- a/src/plain.rs
+++ b/src/plain.rs
@@ -105,10 +105,10 @@ fn color_item_token(token: &Token, bg: Option<Color>) -> ANSIString<'_> {
     };
     #[allow(clippy::match_same_arms)]
     match token {
-        Token::Symbol(text) => style(Color::White.into(), text),
+        Token::Symbol(text) => style(Style::default(), text),
         Token::Qualifier(text) => style(Color::Blue.into(), text),
         Token::Kind(text) => style(Color::Blue.bold(), text),
-        Token::Whitespace => style(Color::White.into(), " "),
+        Token::Whitespace => style(Style::default(), " "),
         Token::Identifier(text) => style(Color::Cyan.into(), text),
         Token::Self_(text) => style(Color::Blue.into(), text),
         Token::Function(text) => style(Color::Yellow.into(), text),
--- a/src/plain.rs
+++ b/src/plain.rs
@@ -2,7 +2,8 @@ use std::io::{Result, Write};
 
 use public_api::{diff::PublicItemsDiff, PublicItem};
 
-use ansi_term::Color;
+use ansi_term::{ANSIString, ANSIStrings, Color, Style};
+use public_api::tokens::*;
 
 use crate::{
     output_formatter::{print_items_with_header, OutputFormatter},
@@ -12,9 +13,13 @@ use crate::{
 pub struct Plain;
 
 impl OutputFormatter for Plain {
-    fn print_items(&self, w: &mut dyn Write, _args: &Args, items: Vec<PublicItem>) -> Result<()> {
+    fn print_items(&self, w: &mut dyn Write, args: &Args, items: Vec<PublicItem>) -> Result<()> {
         for item in items {
-            writeln!(w, "{}", item)?;
+            if args.color.active() {
+                writeln!(w, "{}", color_item(&item))?;
+            } else {
+                writeln!(w, "{}", item)?;
+            }
         }
 
         Ok(())
@@ -30,7 +35,7 @@ impl OutputFormatter for Plain {
             &diff.removed,
             |w, item| {
                 if use_color {
-                    writeln!(w, "{}", Color::Red.paint(item.to_string()))
+                    writeln!(w, "{}", color_item(item))
                 } else {
                     writeln!(w, "-{}", item)
                 }
@@ -47,8 +52,8 @@ impl OutputFormatter for Plain {
                     writeln!(
                         w,
                         "{}\n{}",
-                        Color::Red.paint(changed_item.old.to_string()),
-                        Color::Green.paint(changed_item.new.to_string())
+                        color_item(&changed_item.old),
+                        color_item(&changed_item.new)
                     )
                 } else {
                     writeln!(w, "-{}\n+{}", changed_item.old, changed_item.new)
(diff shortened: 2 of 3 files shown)


douweschulte:
From what I can see now the failing tests are all added features (changed color). But if you @Enselic can confirm that that would be great. 
Sidenote I added the clippy rules to `main.rs` (I previously did the same in `public-api`) so that my editor automatically runs clippy and screams at me when something is not perfect.

Enselic:
Very cool! Love it :)

So, now we enter dangerous territory where a lot of things comes down to personal preference, namely what colors to use :)

I think we might be able to agree on a couple of general rules though:
* **A.** The color scheme should look familiar to most users of our tool. In practice I think this means "similar to the default color schemes (light and dark) in VS Code" since VS Code seems to be the most common IDE for Rust developers.
* **B.** We need the output to work on both light and dark terminal backgrounds. I.e. we should look similar to "Light+ (default light)" and "Dark+ (default dark) themes in VS Code.

Deriving some conclusions from the above rules:
From **A.** we can deduce for example
* `mut` and `impl` should have the same color as `self`

From **B.** we can deduce for example:
* We can't use white for `&` because then it is barely visible on light backgrounds. The way we solve this in `bat` with its `ansi` theme is to "set the color to the foreground color" by "reset all colors". I think we need to do something similar here.

As a next step I propose that I implement at least the above changes and then show you screenshots, and then you can tell me if you think the changes were an overall improvement.

Edit: I am open to deviate from VS Code coloring for "obvious improvements", like maybe having generics in a special color, like you have. Let's try different approaches. 

### Remark on theming

We have not discussed theming yet, but I just wanted to mention that I think it is by far more important to have a good set of default colors rather than the ability for users to customize the colors. In general, users are lazy, and want something that works great out of the box. I am actually quite sceptical of adding support for different themes to this tool. I think it is better to have an uncontroversial and familiar default colors. Theming adds such a large amounts of complexity and discussion, something I don't think this tool needs. It is just a simple API lister and differ after all.

Enselic:
I thought we'd take this one step at a time so we can agree or disagree on each step individually. I just pushed a commit that makes the output look better on both both a dark and a light background, by using the default (empty) style instad of white color.

Before 1cdfca6:
dark-before
light-before

After 1cdfca6:
dark-after
light-after

For the convenience of both you and me I pushed it directly to the PR. Feel very free to force-push it away with your own stuff if you please. But I think we both agree the output needs to work with both light and dark terminal backgrounds. There are more tweaks to be made of course, but I think this is a good first step.

As usual, feel free to come with any kind of feedback, and we'll figure something out.

douweschulte:
I agree with all points. I tried to follow `dark+` in VS code, but I did not check thoroughly so feel free to tweak on some details. The lack of colour instead of white seems very nice. And the theming I do agree with, having a good default is much more important than having a user serviceable colouring. And the way it is handled now, with the 'default' terminal colours it should adapt anyway to the user preferences.

I am fine with pushing to the PR, in that way it is very easy to see what exactly you contributed.

Enselic:
Great 👍 I will keep doing tweaks. I have been doing a lot of tweaks and fixes to `public-api` today, partly tweaks that coloring makes use of. I just pushed a commit now to this PR so that `mut self` becomes one color like in VS Code.

Before 18d67fd:
Skärmavbild 2022-04-17 kl  21 37 22

After  18d67fd:
Skärmavbild 2022-04-17 kl  21 37 31

Also, `impl` and `where` becomes the same color as e.g. `fn`, like in VS Code.

Enselic:
I am happy with what I have now, at least happy enough with the colors to consider the colors releasable. With https://github.com/Enselic/public-api/pull/78 and https://github.com/douweschulte/cargo-public-api/pull/1 I get this with a dark background on macOS with the default Terminal app (that only supports 4-bit colors) with default system colors:

<details>
  <summary>Click to expand</summary>

Skärmavbild 2022-04-18 kl  22 00 10
  
</details>


Same output with a light background:

<details>
  <summary>Click to expand</summary>

Skärmavbild 2022-04-18 kl  22 00 20

  
</details>

A couple of remarks:
* Now types are always green, no matter in what context and what "kind" i.e. primitive or non-primitive. This is the same appearance as in VS Code.
* No text is bold, because I find that distracting. I keep thinking "why is some parts bold and some not bold", which adds mental overhead. At least for my brain. I admit this is in many regards a personal preference :) But VS Code actually has no different in text boldness I think...

If you think this output look good, then I will make a `public-api` release, and then we can continue to iterate on this PR (there are still some things to discuss outside of choosing colors).

douweschulte:
Yes looks great! I agree with both remarks and a new release.

Enselic:
I added + and - to diff output for now because without them the output becomes too confusing. At least when printing changes, and for symmetry with `--color=never` I added them everywhere. I am very open to discuss that though, if you dislike it.

One easy way to test the diff output is to do:

```
cargo run -- --diff-rustdoc-json ~/src/public-api/tests/rustdoc-json/example_api-v0.1.0.json ~/src/public-api/tests/rustdoc-json/example_api-v0.2.0.json
```

which now prints:

Skärmavbild 2022-04-19 kl  21 09 56

I also fixed the test and added a new one.

So personally I consider this Approved. As soon as you also are OK with the current state of this PR, I will merge it and make a v0.9.3 release with it :)

I hope you didn't mind me taking this PR to the finish line. Personally I liked this way of working. But as always, I want you to have fun too, so let me know if I can help with that somehow. BTW, I have to say: Working with render.rs was much more pleasant to work with than `D` and `Display`. I will forever be grateful for that architectural contribution of your to these projects :)

Another BTW: Are you still interested in becoming a formal co-maintainer of public-api and cargo-public-api? I would love to have a co-maintainer on board :)

douweschulte:
I like the `+`/`-`, I did not add them to keep the tests from failing. I am glad you took the PR to the finish line. And I would still be honoured to be a co-maintainer.

BTW I am happy that the `render.rs` worked out 😄.

So to be nicely formal, I also Approve this PR ;-).

Enselic:
Awesome! I invited you as collaborator on both repos. I wrote some guidelines for maintainers here: https://github.com/Enselic/public-api/blob/main/doc/development.md#maintainer-guidelines. Note that guideline **C.** is not implemented yet, but that is at the top of my TODO currently.

I think it goes without saying that as a general rule it is good to ask for code review before merge, but at the same time I don't want to hold you back from making "obvious improvements". Please use good judgement :)

With regards to this PR: I will make a release of this as soon as I can, but I would like to finally setup a CI/CD pipeline that can be used to make releases, so that you can also make releases later. Right now the process is too error prone and has too many steps, and I don't trust myself to never screw up on those. So I think now is a good time to automate releases.

Enselic:
Automating releases in a safe and sound way was was trickier than I anticipated, so I went ahead and released this manually as v0.9.3 (🎉 )

PR #48, opened 2022-05-27
Remove `render_id()` to fix panic with `diesel`

Enselic:
### Step-by-step

1. git checkout and cd into https://github.com/diesel-rs/diesel/tree/master/diesel
2. Until https://github.com/rust-lang/rust/issues/93588 is fixed: apply https://github.com/Enselic/diesel/pull/1/files
3. Run `cargo public-api`

### Expected result
The public API of `diesel` is printed

### Actual result
```
thread 'main' panicked at 'no entry found for key', public-api/src/render.rs:161:5
```

Removing the code in question does not fail any test in e.g. `comprehensive_api`. Maybe the test we have is a purely artificial case? If so, it should be fine to remove the test and the related code, so that we stop panicing on `diesel` rustdoc JSON.

### DIFF
--- a/public-api/src/render.rs
+++ b/public-api/src/render.rs
@@ -3,7 +3,7 @@ use std::rc::Rc;
 
 use rustdoc_types::{
     Abi, Constant, Crate, FnDecl, GenericArg, GenericArgs, GenericBound, GenericParamDef,
-    GenericParamDefKind, Generics, Header, Id, ItemEnum, MacroKind, Term, Type, TypeBinding,
+    GenericParamDefKind, Generics, Header, ItemEnum, MacroKind, Term, Type, TypeBinding,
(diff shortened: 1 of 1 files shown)


Enselic:
@douweschulte Do you remember why you had that code in the first place? I suspect it is needed in some case I can't think of and that we don't cover with `comprehensive_api` tests?

douweschulte:
I am not really sure anymore, but if all tests run correctly now the deletion is justified. I assume this was at the time when I was learning the codebase and the doc_json output, so it could be that I assumed something could be missing when in practice it is always in place.

douweschulte (approved):

Enselic:
Thanks for review! I'll go ahead and merge this then (and then make a release with this fix).

PR #49, opened 2022-05-31
Highlight changes on a Token level in coloured diff output

### LINKED ISSUE #11: Highlight changes in changed API items more clearly

Fraser999:
A small PR to resolve #11.

No problem if you prefer a different approach - I was mainly just getting to know the code :)


cargo-public-api/src/plain.rs
```diff
@@ -49,11 +48,14 @@ impl OutputFormatter for Plain {
             &diff.changed,
             |w, changed_item| {
                 if use_color {
+                    let diff_chars =
+                        diff::chars(&changed_item.old.to_string(), &changed_item.new.to_string());
```
Enselic: Thanks a lot for this PR! I tried it out and it is really cool! I will take a more detailed look on the code as soon as I get time. 

I have one question already: Does `diff` support doing the diff on a `Token` level? I think that would result in an even better diff, e.g. in:

Skärmavbild 2022-05-31 kl  06 45 31

But the diff output formatting is already much better than before, so if not, I am totally fine with doing it on a character-in-a-string level as a first step.

Fraser999: > Thanks a lot for this PR!

No probs

> Does diff support doing the diff on a Token level?

Yes, should be pretty straightforward - I can do that tonight.  I guess it shouldn't be configurable? (Seems a bit overkill to offer that as a new command line option, but I can do that if you prefer).

Enselic: Awesome! You definitely don't need to make it configurable.

cargo-public-api/src/plain.rs
```diff
@@ -49,11 +47,14 @@ impl OutputFormatter for Plain {
...
```
Enselic: Thank you for the update. I haven't used the `diff` crate myself before, but looks like you possibly can use [`diff::slice`](https://docs.rs/diff/latest/diff/fn.slice.html) directly on `Token`s, i.e. without worrying about chars at all? If that would work I think the code could be simplified a lot.

Note to self: Maybe I can replace all/most of my manual diffing algorithm in diff.rs with `diff::slice` on `PublicItem`s 🤔

Fraser999: Good call.  I also haven't used the `diff` crate before - I shoulda RTFM.  Not like it's a big FM :)

The output is similar to the previous attempt.  There are apparently a few more places where the char diff yields a longer unbroken chain, which is maybe slightly more visually appealing.  But I think this simpler implementation is worth that debatable cost.

cargo-public-api/src/plain.rs
```diff
@@ -49,11 +47,16 @@ impl OutputFormatter for Plain {
             &diff.changed,
             |w, changed_item| {
                 if use_color {
+                    let old_tokens: Vec<&str> =
+                        changed_item.old.tokens().map(Token::text).collect();
+                    let new_tokens: Vec<&str> =
+                        changed_item.new.tokens().map(Token::text).collect();
+                    let diff_slice = diff::slice(old_tokens.as_slice(), new_tokens.as_slice());
```
Enselic: Seems like you still mess around with strings a bit more than I would like :)

I'm thinking more along the lines of:
```suggestion
                    let diff_slice = diff::slice(
                        &changed_item.old.tokens().collect::<Vec<_>>(),
                        &changed_item.new.tokens().collect::<Vec<_>>(),
                    );
```
and then make necessary adjustments to the rest of the code from that. The above change makes the type of `diff_slice` be `Vec<Result<&&Token>>` instead of `Vec<Result<&&str>>`.

(BTW, if you happen to be interested in Rust productivity tips, I can recommend VS Code + rust-analyzer, because then you just hold down control+option to temporarily see the type of all vars etc in the code.)

cargo-public-api/src/plain.rs
```diff
@@ -116,3 +117,47 @@ fn color_item_token(token: &Token, bg: Option<Color>) -> ANSIString<'_> {
         Token::Type(text) => style(Color::Green.into(), text),
     }
 }
+
+/// Returns a styled string similar to `color_item_token`, but where whole tokens are highlighted if
+/// they contain a difference.
+fn color_item_with_diff(
+    item: &PublicItem,
```
Enselic: Intuitively it seems like the `Token`s should be enough, and we shouldn't need to .`zip()` on `item.tokens()`?

cargo-public-api/src/plain.rs
```diff
@@ -116,3 +117,47 @@ fn color_item_token(token: &Token, bg: Option<Color>) -> ANSIString<'_> {
...
+) -> String {
+    let diff_iter = diff_slice
+        .iter()
+        .filter_map(|diff_result| match diff_result {
+            diff::Result::Left(&token) => {
+                if is_old_item {
+                    let style = Some(Color::Fixed(9).on(Color::Fixed(52)).bold());
+                    Some((style, token))
+                } else {
+                    None
```
Enselic: Can't we do what `color_item_token(token, None)` does here somehow, instead of doing a `.zip()` later?

Enselic (approved):
Thanks! I did some tests and this seems to work well.

I'm setting Approved on this because I think this PR clearly adds value to the project as it is.

However, if you are interested, I'm pretty sure we can simplify the code significantly by doing everything in one pass instead of doing a `.zip()`, but I haven't figured it out in detail. It just seems like it should be possible, perhaps with some minor refactoring of `color_item_token()`. I added two small comments to pinpoint what I mean.

I want to stress again that you don't need to try to do this simplification. I am willing to merge this PR as is.

@douweschulte I intend to merge this PR (possibly after some further simplification, see above), but before I do I just wanted to ping you in to give you a chance to give input. I know you have looked at this problem too in the past.

cargo-public-api/src/plain.rs
```diff
@@ -116,3 +117,47 @@ fn color_item_token(token: &Token, bg: Option<Color>) -> ANSIString<'_> {
...
+    item: &PublicItem,
+    diff_slice: &[diff::Result<&&Token>],
+    is_old_item: bool,
+) -> String {
+    let diff_iter = diff_slice
+        .iter()
+        .filter_map(|diff_result| match diff_result {
+            diff::Result::Left(&token) => {
+                if is_old_item {
+                    let style = Some(Color::Fixed(9).on(Color::Fixed(52)).bold());
```
douweschulte: I would propose to use the build in colours where possible (eg `Color::Green`) because these are set by the theme of the terminal of the user, so they will always be legible and will fit better in the output of other applications.

cargo-public-api/src/plain.rs
```diff
@@ -116,3 +117,47 @@ fn color_item_token(token: &Token, bg: Option<Color>) -> ANSIString<'_> {
...
+                    Some((style, token))
+                } else {
+                    None
+                }
+            }
+            diff::Result::Both(&token, _) => Some((None, token)),
+            diff::Result::Right(&token) => {
+                if is_old_item {
```
douweschulte: Proposal: you could write:
```rust
(!is_old_item).then(|| (Color::Fixed(10).on(Color::Fixed(22)).bold(), token))
```
Which would compact lines 141-146. But this maybe is a bit terse, feel free to make your own call.

Note: changing the bool to `is_new_item` would make it bit more readable.
Edit: I just spotted that `then_some` is still experimental, but `then` also works.
See: https://doc.rust-lang.org/std/primitive.bool.html#method.then.

douweschulte (approved):
Nice work! I left some style comments, feel free to handle them how you want. Overall a nice addition!

cargo-public-api/src/plain.rs
Fraser999: Yes, it was.  I liked having the `debug_assert_eq` to make sure the diff sequence was valid, but it was probably overkill.  I removed it now.

cargo-public-api/src/plain.rs
Fraser999: I'm not sure I see a benefit to that given the `zip()` is gone now.  I'll leave as is and let you update in a follow-up PR if you don't mind?

cargo-public-api/src/plain.rs
Fraser999: I did try with the defaults, but they produce less legible output - at least for me! :)

cargo-public-api/src/plain.rs
Fraser999: I did make the change as suggested since I quite like `bool::then`.  It's not to everyone's taste though, but it also requires Rust 1.50.0 or greater.  I guess that's not a problem here.

cargo-public-api/src/plain.rs
Enselic: Getting rid of the `.zip()` was what the comment was about, so you already did what I wanted :)

cargo-public-api/src/plain.rs
Enselic: Up until this PR we have only used 3-bit colors which is very widely supported from what I understand. With this PR, we now rely on 256-color support from terminals.

But, I use a very vanilla Terminal.app setup on macOS which AFAIK is the most primitive of all terminal emulators currently is broad use across all major OSes. And the diff looked good for me, and I don't think we can get a good colored diff with only 4-bit colors. So let's go for this and cross our fingers that it works everywhere. Worst case is that we have to tweak the colors later. So the worst case is not horribly bad :)

cargo-public-api/src/plain.rs
Enselic: Yeah Rust 1.50 is more than OK. Our current MSRV is latest stable. (See https://github.com/Enselic/cargo-public-api/blob/main/doc/development.md#minimum-required-rust-version.) That policy might change in the future of course, but so far the current policy has not caused any major inconveniences.

Enselic:
Wow that made such a big difference. The code looks  much better now. Let's merge this!

Next, I will see if I can rewrite diff.rs with the `diff` crate to both get less code and to solve #50. Hopefully should not take many days. Right after that, I plan to release the main branch as v0.11.5.

Fraser999:
Thanks for all the feedback and merging :)

Issue #50, opened 2022-05-31
Diff sometimes unnecessarily big, caused by a off-by-one problem of sorts


Enselic:
## Step-by-step

1. Run `cargo public-api --diff-git-checkouts v0.20.0 v0.21.0` in https://github.com/sharkdp/bat

## Expected result

This diff is printed:
```
Removed items from the public API
=================================
(none)

Changed items in the public API
===============================
-pub enum variant bat::error::Error::SyntectError(::syntect::LoadingError)
+pub enum variant bat::error::Error::SyntectError(::syntect::Error)

Added items to the public API
=============================
+pub enum variant bat::error::Error::SyntectLoadingError(::syntect::LoadingError)
+pub enum variant bat::style::StyleComponent::Default
+pub fn bat::PrettyPrinter::show_nonprintable(&mut self, yes: bool) -> &mut Self
+pub fn bat::error::Error::from(source: ::syntect::Error) -> Self
```

## Actual result
This diff is printed
```
Removed items from the public API
=================================
(none)

Changed items in the public API
===============================
-pub enum variant bat::error::Error::SyntectError(::syntect::LoadingError)
+pub enum variant bat::error::Error::SyntectError(::syntect::Error)
-pub fn bat::error::Error::from(s: &'static str) -> Self
+pub fn bat::error::Error::from(s: String) -> Self
-pub fn bat::error::Error::from(s: String) -> Self
+pub fn bat::error::Error::from(source: ::globset::Error) -> Self
-pub fn bat::error::Error::from(source: ::globset::Error) -> Self
+pub fn bat::error::Error::from(source: ::serde_yaml::Error) -> Self
-pub fn bat::error::Error::from(source: ::serde_yaml::Error) -> Self
+pub fn bat::error::Error::from(source: ::std::io::Error) -> Self
-pub fn bat::error::Error::from(source: ::std::io::Error) -> Self
+pub fn bat::error::Error::from(source: ::std::num::ParseIntError) -> Self
-pub fn bat::error::Error::from(source: ::std::num::ParseIntError) -> Self
+pub fn bat::error::Error::from(source: ::syntect::Error) -> Self

Added items to the public API
=============================
+pub enum variant bat::error::Error::SyntectLoadingError(::syntect::LoadingError)
+pub enum variant bat::style::StyleComponent::Default
+pub fn bat::PrettyPrinter::show_nonprintable(&mut self, yes: bool) -> &mut Self
[...]

PR #52, opened 2022-06-04
Auto-derive `Eq` and `PartialEq` for `PublicItem`

Enselic:
The explicit implementation is identical to the derived one, and by
switching to a derived impl for these we can later make `PublicItem`
derive `Hash` without triggering
https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq.

I'm playing around with fixing #50 with [`hashbag`](https://docs.rs/hashbag/latest/hashbag/) which requires `Hash`. Even if we end up not needing `Hash` later, I think this is a sensible simplifcation.

### DIFF
--- a/public-api/src/item_iterator.rs
+++ b/public-api/src/item_iterator.rs
@@ -177,7 +177,7 @@ fn intermediate_public_item_to_public_item(
 /// of the public API of a crate. Implements [`Display`] so it can be printed. It
 /// also implements [`Ord`], but how items are ordered are not stable yet, and
 /// will change in later versions.
-#[derive(Clone)]
+#[derive(Clone, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


Enselic:
Here is how this change affects the public API of public-api BTW, and it looks reasonable to me:

```
% cargo public-api --output-format markdown --diff-git-checkouts origin/main public-item-derive-eq
```

## Removed items from the public API
(none)

## Changed items in the public API
* `pub fn public_api::PublicItem::eq(&self, other: &Self) -> bool` changed to
  `pub fn public_api::PublicItem::eq(&self, other: &PublicItem) -> bool`

## Added items to the public API
* `pub fn public_api::PublicItem::ne(&self, other: &PublicItem) -> bool`

PR #54, opened 2022-06-06
Simplify `render_sequence()` implementation and invocations

Enselic:
I looked into if it would make sense for us to start depending on `itertools`, more specifically `intersperse()`. It was not a good fit as far as I could tell, but I ended up simplifying (at least to my eyes) the `render_sequence()` code as part of investigating this.

This PR contains those simplifications/clarifications.

### DIFF
--- a/public-api/src/render.rs
+++ b/public-api/src/render.rs
@@ -188,6 +188,28 @@ fn render_sequence<T>(
     start: Vec<Token>,
     end: Vec<Token>,
     between: Vec<Token>,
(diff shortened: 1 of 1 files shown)


Enselic:
@douweschulte Not sure if you are interested in reviewing this, but I figured it wouldn't hurt to give you a chance to do so before I merge, since `render_sequence()` is fully a contribution from you.

douweschulte (approved):
Nice simplifications!

PR #65, opened 2022-06-27
Enforce clippy lints only in CI

Enselic:
Otherwise it becomes unnecessarily tricky to compile older versions of the project with newer versions of the Rust toolchain (for example when doing `git bisect` in the future), because new lints are constantly added.

Keep `#![deny(missing_docs)]` though because
* We only want to do that for the lib
* It seems very unlikely that future version of the compiler will find
    more undocumented public items

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,11 +33,11 @@ jobs:
       - run: RUSTDOCFLAGS='--deny warnings' cargo doc --locked --no-deps
 
   cargo-clippy:
-    name: cargo clippy -- --deny clippy::all --deny clippy::pedantic
+    name: cargo clippy -- --deny clippy::all --deny clippy::pedantic ...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: cargo clippy --locked --all-targets --all-features -- --deny clippy::all --deny clippy::pedantic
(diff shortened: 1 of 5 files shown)


Enselic:
Are you OK with this change @douweschulte? I wanted to check since you added `#![deny(clippy::all, clippy::pedantic)]` in b6c552cf53.

I experienced the downside of in-code deny when I bisected diesel to find https://github.com/rust-lang/rust/issues/98547. I think it is safe to say that cargo-public-api will become annoying to git bisect in the future if we keep denying clippy lints via attributes.  And I want the future to be bright :)

A more elaborate argument against deny in-code is made here: https://github.com/rust-unofficial/patterns/blob/main/anti_patterns/deny-warnings.md

Running `./scripts/run-ci-locally.sh` locally before push I think is less of an inconvenience than problems to build older versions of the code in the future (and running that script should be done before push anyway)

douweschulte (approved):
If you have a use case where this is the wrong way around I fully agree on removing it. Sidenote: I mainly like it because this way of setting it up makes it so my IDE always flags all this stuff, meaning that I have far fewer pushes resulting in some lint failing.

Enselic:
Thanks for review! There probably/hopefully is a way to make the IDE point out clippy problems in a separate config. Personally I run the "rust: cargo clippy" task from `.vscode/tasks.json` before push. Speaking of, I forgot to update `.vscode/tasks.json` 😅. I will do that before merge.

PR #72, opened 2022-07-10
Add `--deny=all` option to exit with failure if there is an API diff

### LINKED ISSUE #63: Add `--deny` flag for diffs in CI usage situations
Hi there! Cool project 😄 

I'd like to use this as part of a CI pipeline but it doesn't look like there's any way to
programmatically check if something in the public API changed.

RustFmt has a [`--check` flag](https://github.com/rust-lang/rustfmt#checking-style-on-a-ci-server) which exits with `0` if the input would be formatted successfully
and `1` if it wouldn't be.

I think having a similar flag as part of this project would allow it to be used in the context
of CI.

matthiasbeyer:
Closes #63 

This is merely a proposal.

I am not very sure about the actual flag name. `--check` seems to be ambiguous for me, maybe something along `--check-diff-empty` or `--is-empty` is more asppropriate.

Tell me what you think!

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -2,7 +2,9 @@ use std::io::stdout;
 use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, Context, Result};
+use arg_types::{Color, DenyMethod, OutputFormat};
 use output_formatter::OutputFormatter;
+use public_api::diff::PublicItemsDiff;
 use public_api::{
     public_api_from_rustdoc_json_str, Options, PublicItem, MINIMUM_RUSTDOC_JSON_VERSION,
 };
@@ -11,6 +13,7 @@ use clap::Parser;
 use rustdoc_json::BuildError;
 
 mod arg_types;
+mod error;
 mod git_utils;
 mod markdown;
 mod output_formatter;
@@ -71,15 +74,20 @@ pub struct Args {
     #[clap(long, min_values = 2, max_values = 2)]
     diff_rustdoc_json: Option<Vec<String>>,
 
+    /// Exit with failure if the specified API diff is detected. all = deny
+    /// additions, changes, and removals of public items in the API
+    #[clap(long, arg_enum)]
+    deny: Option<Vec<DenyMethod>>,
+
     /// What output format to use. You can select between "plain" and "markdown".
     #[clap(long, name = "FORMAT", default_value = "plain")]
-    output_format: arg_types::OutputFormat,
+    output_format: OutputFormat,
 
     /// Whether or not to use colors. You can select between "auto", "never", "always".
     /// If "auto" (the default), colors will be used if stdout is a terminal. If you pipe
     /// the output to a file, colors will be disabled by default.
     #[clap(long, default_value = "auto")]
-    color: arg_types::Color,
+    color: Color,
 
     /// Show detailed info about processing. For debugging purposes. The output
     /// is not stable and can change across patch versions.
@@ -95,12 +103,34 @@ pub struct Args {
 fn main_() -> Result<()> {
     let args = get_args();
 
-    if let Some(commits) = &args.diff_git_checkouts {
-        print_public_items_diff_between_two_commits(&args, commits)
+    let diff = if let Some(commits) = &args.diff_git_checkouts {
+        Some(print_diff_between_two_commits(&args, commits)?)
     } else if let Some(files) = &args.diff_rustdoc_json {
-        print_public_items_diff_between_two_rustdoc_json_files(&args, files)
+        Some(print_diff_between_two_rustdoc_json_files(&args, files)?)
     } else {
-        print_public_items_of_current_commit(&args)
+        print_public_items_of_current_commit(&args)?;
+        None
+    };
+
+    check_diff(&args, diff)
+}
+
+fn check_diff(args: &Args, diff: Option<PublicItemsDiff>) -> Result<()> {
+    match (&args.deny, diff) {
+        // We were requested to deny diffs, so make sure there is no diff
+        (Some(_deny), Some(diff)) => {
+            if diff.is_empty() {
+                Ok(())
+            } else {
+                Err(anyhow!(error::Error::DiffDenied))
+            }
+        }
+
+        // We were requested to deny diffs, but we did not calculate a diff!
+        (Some(_), None) => Err(anyhow!("`--deny` can only be used when diffing")),
+
+        // No diff related stuff to care about, all is Ok
+        _ => Ok(()),
     }
 }
 
@@ -113,7 +143,7 @@ fn print_public_items_of_current_commit(args: &Args) -> Result<()> {
     Ok(())
 }
 
-fn print_public_items_diff_between_two_commits(args: &Args, commits: &[String]) -> Result<()> {
+fn print_diff_between_two_commits(args: &Args, commits: &[String]) -> Result<PublicItemsDiff> {
     let old_commit = commits.get(0).expect("clap makes sure first commit exist");
     let old = collect_public_items_from_commit(Some(old_commit))?;
 
@@ -123,10 +153,10 @@ fn print_public_items_diff_between_two_commits(args: &Args, commits: &[String]) 
     print_diff(args, old, new)
 }
 
-fn print_public_items_diff_between_two_rustdoc_json_files(
+fn print_diff_between_two_rustdoc_json_files(
     args: &Args,
     files: &[String],
-) -> Result<()> {
+) -> Result<PublicItemsDiff> {
     let options = get_options(args);
 
     let old_file = files.get(0).expect("clap makes sure first file exists");
@@ -138,13 +168,13 @@ fn print_public_items_diff_between_two_rustdoc_json_files(
     print_diff(args, old, new)
 }
 
-fn print_diff(args: &Args, old: Vec<PublicItem>, new: Vec<PublicItem>) -> Result<()> {
-    let diff = public_api::diff::PublicItemsDiff::between(old, new);
+fn print_diff(args: &Args, old: Vec<PublicItem>, new: Vec<PublicItem>) -> Result<PublicItemsDiff> {
+    let diff = PublicItemsDiff::between(old, new);
     args.output_format
         .formatter()
         .print_diff(&mut stdout(), args, &diff)?;
 
-    Ok(())
+    Ok(diff)
 }
 
 /// Get CLI args via `clap` while also handling when we are invoked as a cargo
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -6,6 +6,7 @@
 use std::path::{Path, PathBuf};
 
 use assert_cmd::Command;
+use predicates::str::contains;
 use serial_test::serial;
 
 #[serial]
@@ -54,7 +55,7 @@ fn virtual_manifest_error() {
     cmd.arg(current_dir_and("tests/virtual-manifest/Cargo.toml"));
     cmd.assert()
         .stdout("")
-        .stderr(predicates::str::contains(
+        .stderr(contains(
             "Listing or diffing the public API of an entire workspace is not supported.",
         ))
         .failure();
@@ -88,10 +89,7 @@ fn deny_when_not_diffing() {
     let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
     cmd.arg("--deny=all");
     cmd.assert()
-        .stdout("")
-        .stderr(predicates::str::contains(
-            "`--deny` can only be used with `--diff-git-checkouts`",
-        ))
+        .stderr(contains("`--deny` can only be used when diffing"))
         .failure();
 }
 
@@ -120,7 +118,9 @@ fn deny_with_diff() {
     cmd.arg("v0.0.4");
     cmd.arg("v0.0.5");
     cmd.arg("--deny=all");
-    cmd.assert().stderr("").failure();
+    cmd.assert()
+        .stderr(contains("The API diff is not allowed as per --deny"))
+        .failure();
 }
 
 #[serial]
@@ -281,7 +281,7 @@ fn diff_public_items_missing_one_arg() {
     cmd.arg("--diff-git-checkouts");
     cmd.arg("v0.2.0");
     cmd.assert()
-        .stderr(predicates::str::contains(
+        .stderr(contains(
             "requires at least 2 values but only 1 was provided",
         ))
         .failure();
@@ -308,9 +308,7 @@ fn list_public_items_markdown() {
 fn verbose() {
     let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
     cmd.arg("--verbose");
-    cmd.assert()
-        .stdout(predicates::str::contains("Processing \""))
-        .success();
+    cmd.assert().stdout(contains("Processing \"")).success();
 }
 
 #[test]
@@ -339,9 +337,9 @@ fn assert_presence_of_own_library_items(mut cmd: Command) {
 
 fn assert_presence_of_args_in_help(mut cmd: Command) {
     cmd.assert()
-        .stdout(predicates::str::contains("--with-blanket-implementations"))
-        .stdout(predicates::str::contains("--manifest-path"))
-        .stdout(predicates::str::contains("--diff-git-checkouts"))
+        .stdout(contains("--with-blanket-implementations"))
+        .stdout(contains("--manifest-path"))
+        .stdout(contains("--diff-git-checkouts"))
         .success();
 }
 
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -82,6 +82,21 @@ fn diff_public_items() {
 
 #[serial]
 #[test]
+fn deny_when_not_diffing() {
+    ensure_test_crate_is_cloned(); // Because we still list the API
+
+    let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
+    cmd.arg("--deny=all");
+    cmd.assert()
+        .stdout("")
+        .stderr(predicates::str::contains(
+            "`--deny` can only be used with `--diff-git-checkouts`",
+        ))
+        .failure();
+}
+
+#[serial]
+#[test]
 fn diff_public_items_with_manifest_path() {
     ensure_test_crate_is_cloned();
 
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -125,6 +125,22 @@ fn deny_with_diff() {
 
 #[serial]
 #[test]
+fn deny_with_invalid_arg() {
+    ensure_test_crate_is_cloned();
+
+    let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
+    cmd.current_dir(test_crate_path());
+    cmd.arg("--diff-git-checkouts");
+    cmd.arg("v0.0.4");
+    cmd.arg("v0.0.5");
+    cmd.arg("--deny=invalid");
+    cmd.assert()
+        .stderr(contains("\"invalid\" isn't a valid value"))
+        .failure();
+}
+
+#[serial]
+#[test]
 fn diff_public_items_with_manifest_path() {
     ensure_test_crate_is_cloned();
 
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -111,6 +111,20 @@ fn deny_without_diff() {
 
 #[serial]
 #[test]
+fn deny_with_diff() {
+    ensure_test_crate_is_cloned();
+
+    let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
+    cmd.current_dir(test_crate_path());
+    cmd.arg("--diff-git-checkouts");
+    cmd.arg("v0.0.4");
+    cmd.arg("v0.0.5");
+    cmd.arg("--deny=all");
+    cmd.assert().stderr("").failure();
+}
+
+#[serial]
+#[test]
 fn diff_public_items_with_manifest_path() {
     ensure_test_crate_is_cloned();
 
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -97,6 +97,20 @@ fn deny_when_not_diffing() {
 
 #[serial]
 #[test]
+fn deny_without_diff() {
+    ensure_test_crate_is_cloned();
+
+    let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
+    cmd.current_dir(test_crate_path());
+    cmd.arg("--diff-git-checkouts");
+    cmd.arg("v0.2.0");
+    cmd.arg("v0.3.0");
+    cmd.arg("--deny=all");
+    cmd.assert().success();
+}
+
+#[serial]
+#[test]
 fn diff_public_items_with_manifest_path() {
     ensure_test_crate_is_cloned();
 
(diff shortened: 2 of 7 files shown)


matthiasbeyer:
Not sure why CI fails, AFAICS, `std::process::ExitCode` implements `Termination`... ?

matthiasbeyer:
Hah, fun! `Result` does not implement `Termination` for `T: Termination`. I guess we have to work around that then, because I suspect returning `Result` from main is nice because of the error printing automatism.

cargo-public-api/src/main.rs
```diff
@@ -85,18 +85,30 @@ pub struct Args {
...
 
-fn main_() -> Result<()> {
+fn main_() -> Result<bool> {
     let args = get_args();
 
-    if let Some(commits) = &args.diff_git_checkouts {
-        print_public_items_diff_between_two_commits(&args, commits)
+    let exit_success = if let Some(commits) = &args.diff_git_checkouts {
+        let there_is_diff = print_public_items_diff_between_two_commits(&args, commits)?;
+        args.check && !there_is_diff
```
Enselic: CI fails because this returns `false`  even without `args.check` being true, which makes the process exit with failure. But if `args.check` is not true, diffing should always make the app exit with success.

I wonder if the code wouldn't become nicer if we instead of duplicating the code for `there_is_diff` handled this centrally? Since `args` is passed down, you can know deeper down if `args.check` is true or not.

I *think* I would be willing to call `std::process::exit(1)` deep down if that greatly simplifies the code. It is actually  more important that we write tests for what behaviour we want than exactly how we implement it. I sketched out what kind of tests I think we need in https://github.com/Enselic/cargo-public-api/issues/63#issuecomment-1167111370

Enselic (changes requested):
Thanks a lot for working on this! 🎉 

I made a high-level comment. I will wait with a detailed review until we have the big strokes in place.

Regarding if we should call the flag `--check`: I agree it does not feel like a super good fit. On the upside, it is the same name used in e.g. `rustfmt --check`. But I think it makes more semantic sense in there. `rustfmt --check` can be read as "check if formatting is OK".

In our case, `cargo public-api --diff-git-checkouts v0.1.0 v0.2.0 --check` would read as "check that the diff is OK between v0.1.0 and v0.2.0". That doesn't seem to make much sense.  Maybe call the flag `--deny-changes`? Besides, in the future, we might want to allow diffs to contain only added items but not removed items. In other words, allow users to have higher resolution on what diff that is allowed. Not sure what to call such an option though.

Another high level comment I have: It would be great if we could add tests to the code before we polish the implementation. Let me know if you want help with adding tests apart from the sketch I made in the comment I linked to.

cargo-public-api/src/main.rs
matthiasbeyer: > CI fails because this returns false even without args.check being true, which makes the process exit with failure. But if args.check is not true, diffing should always make the app exit with success.

Yeah, I added a patch simplifying the implementation, so that it is clearly visible what is happening.

matthiasbeyer:
>  Maybe call the flag --deny-changes? Besides, in the future, we might want to allow diffs to contain only added items but not removed items. In other words, allow users to have higher resolution on what diff that is allowed. Not sure what to call such an option though.

Yes, that's why I am particularly unhappy with `--check` :laughing: 

If we think about what a user might want:

* Check whether there is any diff
* Check whether there were things added
* Check whether there were things removed
* Check whether there were things changed

So I guess the flags should reflect that, whereas "check" would mean that they want a "successfull" exit if nothing is added/removed/changed - at least that's what I would want.

Maybe something like

* `--deny=all` (with the `=all` part being optional)
* `--deny=added`
* `--deny=removed`
* `--deny=changed`

matthiasbeyer:
Forgot to add:

>  It would be great if we could add tests to the code before we polish the implementation. Let me know if you want help with adding tests apart from the sketch I made in the comment I linked to.

Yes, definitively. Once we're settled I would like to add tests, and once we're settled I would also like to implement the other flags (so `--deny=all` but also `--deny=<all others>` :smile: ).

Enselic:
`--deny=all` or `--deny=added,removed` etc sounds good to me, at least as a first iteration!

Also feel free to only do `--deny=all` in this first PR. Not sure I think we should allow only `--deny` because that does not make semantic sense IMHO.

BTW your second set of CI failures was caused by an unrelated change to rustup released as 1.25.0 . I have merged a fix for that now via https://github.com/Enselic/cargo-public-api/pull/74 so if you rebase on that CI should become more happy

matthiasbeyer:
Rebased and added the changes we discussed. I would like to implement the `--deny=additions`/etc. variants in this PR as well (if you approve) because right now it does not make that much sense with only `all` as a possible value for `--deny` :laughing:

cargo-public-api/src/main.rs
```diff
@@ -85,18 +85,49 @@ pub struct Args {
...
+//
+// Returns Ok(true) if everything went well
+fn main_() -> Result<bool> {
     let args = get_args();
 
-    if let Some(commits) = &args.diff_git_checkouts {
-        print_public_items_diff_between_two_commits(&args, commits)
+    let exit_success = if let Some(commits) = &args.diff_git_checkouts {
+        let there_is_diff = print_public_items_diff_between_two_commits(&args, commits)?;
+        if args.deny.is_some() {
```
Enselic: I think we can simplify this code (mainly getting rid of duplication), but before we do, we should get tests in place. After we have tests we can iterate quickly on the refactoring. Let me know if you want more help than the pointers I gave an an earlier comment.

Enselic (changes requested):

matthiasbeyer:
Does the above two patches for tests look good to you?

Enselic:
There is an [upcoming change](https://github.com/rust-lang/rust/pull/99287#pullrequestreview-1040717371) that might break this tool, so I need to take care of that with highest prio. I'll get back to this PR as soon as possible after that!

matthiasbeyer:
Sure, take your time! I am not in a hurry and won't go away! :laughing:

cargo-public-api/src/main.rs
```diff
@@ -85,18 +85,49 @@ pub struct Args {
...
+    /// This flag can be used to check whether there is a diff.
+    ///
+    /// If specified with `--deny=all`, no diff is allowed to be there and cargo-public-api exits
+    /// with a nonzero exit code if there is.
+    #[clap(long, arg_enum)]
+    deny: Option<Vec<DenyMethod>>,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, clap::ArgEnum)]
+#[clap(rename_all = "lower")]
```
Enselic: Nice! I didn't know clap supported this. I have done this manually in `arg_types.rs`. That code could probably use this nifty `clap` attribute instead. Not needed to be taken care of this PR though of course.

cargo-public-api/src/main.rs
```diff
@@ -85,18 +85,49 @@ pub struct Args {
...
     rustdoc_json_toolchain: String,
+
+    /// Deny diff
+    ///
+    /// This flag can be used to check whether there is a diff.
+    ///
+    /// If specified with `--deny=all`, no diff is allowed to be there and cargo-public-api exits
+    /// with a nonzero exit code if there is.
+    #[clap(long, arg_enum)]
+    deny: Option<Vec<DenyMethod>>,
```
Enselic: Sorry to nitpick, but can you move this arg to below `diff_rustdoc_json: Option<Vec<String>>,` please? I'd like to keep the `hide = true` ones at the bottom, and this arg belongs in the diff args group.

cargo-public-api/src/main.rs
Enselic: Here is how I approximately would like this code to be:

```rust
    let diff: Option<PublicItemsDiff> = if let Some(commits) = &args.diff_git_checkouts {
        print_public_items_diff_between_two_commits(&args, commits)?;
    } else if let Some(files) = &args.diff_rustdoc_json {
        print_public_items_diff_between_two_rustdoc_json_files(&args, files)?;
    } else {
        print_public_items_of_current_commit(&args)?;
        None
    };

    if let Some(diff) = diff {
        if args.deny.is_some() && diff.is_empty() {
            Err(Error::DiffDenied)?
        } else {
            Ok(())
        }
    } else {
        Ok(())
    }
```

which requires

```rust
#[derive(thiserror::Error, Debug)]
enum Error {
    #[error("THe API diff is not allowed as per --deny")]
    DiffDenied,
}
```

and also 

```toml
thiserror = "1.0.29"
```

in cargo-public-api/Cargo.toml

Note: Not polished, I just wanted to convey the idea.

cargo-public-api/src/main.rs
```diff
@@ -85,18 +85,49 @@ pub struct Args {
...
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, clap::ArgEnum)]
+#[clap(rename_all = "lower")]
+enum DenyMethod {
+    All,
 }
 
-fn main_() -> Result<()> {
```
Enselic, line 95: Rather than changing to `Result<bool>` which I find to be confusing, I think instead we should introduce an error. See other comments for more details.

cargo-public-api/src/main.rs
```diff
@@ -230,6 +261,7 @@ fn main() -> Result<()> {
             }
             _ => Err(e),
         },
-        result => result,
+        Ok(false) => std::process::exit(1),
+        Ok(true) => Ok(()),
```
Enselic: I think this should look something like

```
        Err(e) => match e.root_cause().downcast_ref::<Error>() {
            Some(Error::DiffDenied) => {
                std::process::exit(1)
            }
            _ => Err(e),
        },
        result => result,
```

Note: Not polished code, just showing the general idea.

public-api/src/diff.rs
```diff
@@ -106,6 +106,12 @@ impl PublicItemsDiff {
             added,
         }
     }
+
+    /// Check whether the diff is empty
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
```
Enselic, line 112: Please add a call to this in existing tests in `tests/public-api-lib-tests.rs` and assert on that it returns the right value in all existing tests. (Or at least a representative sub-set.)

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -87,6 +87,28 @@ fn diff_public_items() {
...
+
+    let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
+    cmd.current_dir(test_crate_path());
+    cmd.arg("--deny=all");
+    cmd.arg("--diff-git-checkouts");
+    cmd.arg("v0.2.0");
+    cmd.arg("v0.3.0");
+    cmd.assert().success();
+}
+
```
Enselic: I think we can get away with just three tests, namely these (which I based on your tests, thanks!):

```rust
#[test]
fn test_check_no_diff_fails() {
    let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
    cmd.arg("--deny=all");
    cmd.assert()
        .stdout("")
        .stderr(predicates::str::contains(
            "`--deny` can only be used with `--diff-git-checkouts`",
        ))
        .failure();
}

#[serial]
#[test]
fn test_check_no_diff_succeeds_on_no_diff() {
    ensure_test_crate_is_cloned();

    let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
    cmd.current_dir(test_crate_path());
    cmd.arg("--diff-git-checkouts");
    cmd.arg("v0.2.0");
    cmd.arg("v0.3.0");
    cmd.arg("--deny=all");
    cmd.assert().success();
}

#[serial]
#[test]
fn test_check_no_diff_fails_with_diff() {
    ensure_test_crate_is_cloned();

    let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
    cmd.current_dir(test_crate_path());
    cmd.arg("--diff-git-checkouts");
    cmd.arg("v0.0.4");
    cmd.arg("v0.0.5");
    cmd.arg("--deny=all");
    cmd.assert().stderr("").failure();
}

```

Enselic (dismissed):
Time to revisit this PR! 🎉  I added some more comments now to flesh out what kind of changes I would like to see. Let me know if you have any questions :) Looking forward to the next review round!

Probably a good idea to rebase on latest master BTW, because I refactored some of our tests recently.

To be clear: I am just making proposals, so don't interpret this as how it **must** look. Feel free to come up with alternatives, and we can discuss them :)

matthiasbeyer:
Here comes a rebase with the fixes you wanted (I hope I did not miss anything)!
Feel free to suggest another round of fixes if you want, I am looking forward!

matthiasbeyer:
FWIW, I'm going to mark the discussions I consider done as "completed", feel free to re-open them of course, if you think we can do better.

cargo-public-api/src/main.rs
matthiasbeyer: :+1: nice that I could teach you something :laughing:

cargo-public-api/src/main.rs
matthiasbeyer: Sure. Done.

cargo-public-api/src/main.rs
matthiasbeyer: Jup, done I think.

cargo-public-api/src/main.rs
matthiasbeyer: Done.

cargo-public-api/src/main.rs
matthiasbeyer: I used a `if let Err(e)`, which is a bit more idomatic here I guess.

public-api/src/diff.rs
matthiasbeyer: I think I did it the way you want :smile:

cargo-public-api/tests/cargo-public-api-bin-tests.rs
matthiasbeyer: I think I've done this change to your liking, feel free to nag me about this, of course!

cargo-public-api/src/main.rs
```diff
@@ -92,15 +102,43 @@ pub struct Args {
...
+
+#[derive(Debug, thiserror::Error)]
+enum Error {
+    #[error("The API diff is not allowed as per --deny")]
+    DiffDenied,
+}
+
+// Actual main implementation
+//
+// Returns Ok(true) if everything went well
```
Enselic: This part is outdated now, and I think we can remove the line above too actually

cargo-public-api/src/main.rs
```diff
@@ -71,6 +72,15 @@ pub struct Args {
     #[clap(long, min_values = 2, max_values = 2)]
     diff_rustdoc_json: Option<Vec<String>>,
 
+    /// Deny diff
```
Enselic: This is a bit too terse as this is the only thing told about the arg in "short help" mode (`cargo run -- -h`)

cargo-public-api/src/main.rs
```diff
@@ -92,15 +102,43 @@ pub struct Args {
...
+        )?)
     } else {
-        print_public_items_of_current_commit(&args)
+        print_public_items_of_current_commit(&args)?;
+        None
+    };
+
+    if let Some(diff) = diff {
+        if args.deny.is_some() && diff.is_empty() {
+            anyhow::bail!(Error::DiffDenied)
```
Enselic: `bail!` expands to `return`, but it is enough to just evaluate to what we want to return, since this is the last statement, so let's simplify

cargo-public-api/src/main.rs
```diff
@@ -231,13 +272,14 @@ or
 
 /// Wrapper to handle <https://github.com/rust-lang/rust/issues/46016>
 fn main() -> Result<()> {
-    match main_() {
-        Err(e) => match e.root_cause().downcast_ref::<std::io::Error>() {
+    if let Err(e) = main_() {
```
Enselic: I started to think "maybe we don't need to change this at all", and seems like I was right, because if we leave this unchanged, the tool will simply say "`Error: The API diff is not allowed as per --deny`", which seems like a good thing to do. So let's just leave this code as-is.

Enselic (commented):
Thanks a lot for the update! That is a much better structuring of the code.

I had quite a few changes I would like still, so I figured it's easier for me to just push a commit (which I will do shortly after posting this comment) than making 20 small comments. I hope you don't mind. Do you think my changes look OK? Essentially: can you review my changes please? 😇 

Before merge I will squash commits together and remove my name from the code so that you will still get full credit for the change in git log and releases notes etc. Without you this wouldn't have happened, so you deserve the credit :)

cargo-public-api/src/main.rs
Enselic: (fixed in https://github.com/Enselic/cargo-public-api/pull/72/commits/1ff21def29f5b07de3df91eba9f6691bd28469a9)

cargo-public-api/src/main.rs
Enselic: (fixed in https://github.com/Enselic/cargo-public-api/pull/72/commits/1ff21def29f5b07de3df91eba9f6691bd28469a9)

cargo-public-api/src/main.rs
Enselic: (fixed in https://github.com/Enselic/cargo-public-api/pull/72/commits/1ff21def29f5b07de3df91eba9f6691bd28469a9)

cargo-public-api/src/main.rs
Enselic: (fixed in https://github.com/Enselic/cargo-public-api/pull/72/commits/1ff21def29f5b07de3df91eba9f6691bd28469a9)

Enselic:
(When you do the review I recommend diffing against origin/main. The diff of my commit in isolation is probably confusing.)

cargo-public-api/src/main.rs
```diff
@@ -95,12 +103,34 @@ pub struct Args {
...
+
+fn check_diff(args: &Args, diff: Option<PublicItemsDiff>) -> Result<()> {
+    match (&args.deny, diff) {
+        // We were requested to deny diffs, so make sure there is no diff
+        (Some(_deny), Some(diff)) => {
+            if diff.is_empty() {
+                Ok(())
+            } else {
+                Err(anyhow!(error::Error::DiffDenied))
+            }
```
matthiasbeyer, line 126:
```suggestion
            diff.is_empty().then(|| Ok(())).unwrap_or_else(|| anyhow!(error::Error::DiffDenied))
```

What about this (modulo some linebreaks maybe)? :laughing:

cargo-public-api/src/main.rs
```diff
@@ -95,12 +103,34 @@ pub struct Args {
...
+        }
+
+        // We were requested to deny diffs, but we did not calculate a diff!
+        (Some(_), None) => Err(anyhow!("`--deny` can only be used when diffing")),
```
matthiasbeyer, line 130:
```suggestion
        (Some(_), None) => anyhow::bail!("`--deny` can only be used when diffing"),
```

matthiasbeyer (commented):
I cannot block this PR because I am the author, but I have some minor things that could be improved.

Tell me if you want me to improve them or if you're okay with the current state and we can ignore these minor things.

And: Can I do the squashing please? You helped a great deal and I'd like to honor that in the commit! Just tell me if you only want one commit or whether several are okay!

cargo-public-api/src/main.rs
Enselic: Well, apart from that it is more complicated (`bool.then()` returns an `Option<T>`, but there is no need for that here I think) and does not compile, we could use that instead I guess 😅 

Why not use the current code though? Don't you think it is easier to see what happens? 

If you strongly prefer it to be something else I will for sure consider it, but I'd have to see something that compiles first :) (My deepest apologies if it does compile and I just misunderstood how to apply the change)

But I have to be honest: I strongly prefer simplicity over "smartness" (in lack of a better word).

cargo-public-api/src/main.rs
Enselic: Hmm not sure I'm a fan of that, because `anyhow::bail!` it expands to `return Err(anyhow!("--deny can only be used when diffing"))`, but there is no need to return early (i.e. "bail"), because the evaluation of the expression is already what will be returned.

Enselic:
Thanks a lot for code review! I responded in comments.

> And: Can I do the squashing please? You helped a great deal and I'd like to honor that in the commit! Just tell me if you only want one commit or whether several are okay!

Yes, of course! Feel free!

Normally I do regular git merges of all commits in a PR, but in this case I think there is a bit too much of fixup! and squash! commits to make sense to make them part of the persistent git log of the project

Doing the equivalent of `git rebase --autosquash` would probably be sufficient, but feel free to make any kind of cleanup you wish, as long as you use good judgement :)

cargo-public-api/src/main.rs
matthiasbeyer: > Why not use the current code though? Don't you think it is easier to see what happens?

I just like to explore all options in terms of "idomadicness" :smile: 

I'm okay with the current implementation of course, I consider this resolved.

matthiasbeyer:
Here we go.

I had to split your patch into several smaller ones to be able to squash them into the right commits properly. Next time please commit the individual changes `git add -p`. But no worries, I am some kind of git guru (some say so), so it wasn't a problem for me :laughing: 

I guess this is now ready for a final round of review and IMO also merging.
Make sure to have a look at the commit messages, whether I gave proper credit and the trailers are correct!

Besides that: :tada: 

Was a really pleasure to work with you on this! :heart:

Enselic (approved):
> Next time please commit the individual changes git add -p.

Yeah I'm actually quite good at that normally, but since in this case I planned to squash everything to one commit, I didn't bother :)

The new commit structure certainly looks good enough for merging as is, thanks!

I will let the PR sit for a day or two, because I've found that quite frequently you can think of some minor detail the hours after merging that makes you regret merging. So by waiting a day or two we can make some final adjustments if we think of any. Small things like typos and what not.

Plus, other maintainers gets a chance to review a bit too, if they want.

Yes, thanks for a good collaboration! 🎉 

Edit: Let's merge this now, I would like to push some code that depends on this PR being merged :P

Enselic:
For the record, this PR changes the `public-api` public API in the following way (as expected):


```
% cargo public-api --manifest-path public-api/Cargo.toml --diff-git-checkouts v0.12.3 main
Removed items from the public API
=================================
(none)

Changed items in the public API
===============================
(none)

Added items to the public API
=============================
+pub fn public_api::diff::PublicItemsDiff::is_empty(&self) -> bool
```

I mainly mention this because I like to diff the public api of public-api, because it's so meta. The secondary reason is that it is good to keep track of what PR makes what changes to the public API of public-api :)

Enselic:
Released as https://github.com/Enselic/cargo-public-api/releases/tag/v0.12.4 to crates.io 🥳

matthiasbeyer:
You released new features as a patch release?

Enselic:
In the Rust ecosystem, 0.X.Y is interpreted to mean that the major version is X and the minor version is Y. There is no number for patch versions unless you use 1.y.z in which case z is the patch number.

(Due to the rustdoc JSON format still changing in incompatible ways it is too early to enter 1.0.0 for this project IMHO).

See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio for more info

> This compatibility convention is different from SemVer in the way it treats versions before 1.0.0. While SemVer says there is no compatibility before 1.0.0, Cargo considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.

matthiasbeyer:
Well, you're talking about _compatibility_ here. I'm just talking about patch release vs. minor release. Cargo treats compatibility different than semver suggests, that does not mean that in `x.y.z` if `x` is `0` that `y` is "major". `x` is still major and `y` minor and `z` patch.

I was just wondering why you'd release a new feature with an increase in the "patch" position of the version string instead of a "minor", which would what semver defines. With your reasoning, it is understandable for me though.

Enselic:
If I would have released this as 0.13.0 I would have had to update https://github.com/Enselic/cargo-public-api#compatibility-matrix which I find to be too much of an effort TBH, plus it inflates version numbers unnecessarily (IMHO).

The most important thing for me is that nothing actually breaks. For example, if I remove an item from the public API and then release that without bumping the middle number, people will get build errors when they do `cargo install cargo-public-api` of a version that has not been updated accordingly. That would not be acceptable.

With this release, there should not be any breakage anywhere, which is the most important aspect.

So in practice, I treat the middle number as the major number, and the last number as the minor number. As far as I know that is quite commonly done in the Rust ecosystem.

PR #77, opened 2022-07-17
Inline imports instead of showing `pub use` in APIs

### LINKED ISSUE #76: From `nightly-2022-07-17`, re-exports from private modules show no type/kind
Consider this `src/lib.rs`:
```rust
mod private {
    pub fn f(x: usize) {}
}
pub use private::f;
```

With rustdoc JSON built by `nightly-2022-07-16`, our tooling will give this list of public items:
```
pub fn example_api::f(x: usize)
pub mod example_api
```

With rustdoc JSON built by `nightly-2022-07-17`, our tooling will give this list of public items:
```
pub mod example_api
pub use example_api::f
```

This is problematic for us, because e.g. doing an API diff after changing the parameter type of `f(x: usize)` to something else, the diff will show nothing, since the type is not part of the output any longer. It also deviates from the corresponding `cargo doc` HTML output, which performs inlining.

This was changed upstream by https://github.com/rust-lang/rust/pull/99287 to fix a handful of ICEs, so there is good reason for it.

I think the fix from our side should be to manually inline items re-exported from private modules. I have done some prototyping and it does not seem that hard to do. I do however think rustdoc JSON should expose the parent of an item in a convenient way first (https://github.com/rust-lang/rust/issues/93524). I will try to get that fixed.

Note that we do not need to bump `MINIMUM_RUSTDOC_JSON_VERSION` yet because of this, because we can still parse the output from both toolchains. The output has only changed form, not format (in an incompatible way).

Enselic:
This change was triggered by https://github.com/rust-lang/rust/pull/99287, but it would have made sense to do even without that upstream change.

Closes #76

Note that we lose

```
pub fn comprehensive_api::RngCore::fill_bytes(&mut self, dest: &mut [u8])
pub fn comprehensive_api::RngCore::next_u32(&mut self) -> u32
pub fn comprehensive_api::RngCore::next_u64(&mut self) -> u64
pub fn comprehensive_api::RngCore::try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Error>
```

from the listed public API of `comprehensive_api`, and this seems like a [bug](https://github.com/rust-lang/rust/pull/99287#issuecomment-1186589796), but I do not consider it serious enough to warrant some kind of emergency at this moment.

This is currently blocked on https://github.com/aDotInTheVoid/rustdoc-types/pull/11, but I wanted to prepare this PR anyway so me and other have time to think about the implications.

Our nightly CI is currently failing, but we just have to live with that for a couple of days. Or longer we are unlucky.

### DIFF
--- a/public-api/src/intermediate_public_item.rs
+++ b/public-api/src/intermediate_public_item.rs
@@ -1,7 +1,7 @@
 use crate::render;
 use std::rc::Rc;
 
-use rustdoc_types::{Crate, Item, ItemEnum};
+use rustdoc_types::{Crate, Item};
 
 use crate::tokens::Token;
 
@@ -13,6 +13,11 @@ use crate::tokens::Token;
 pub struct IntermediatePublicItem<'a> {
     /// The item we are effectively wrapping.
     pub item: &'a Item,
+
+    /// The name of the item. Normally this is [Item::name]. But in the case of
+    /// renamed imports (`pub use other::item as foo;`) it is the new name.
+    pub name: &'a str,
+
     pub root: &'a Crate,
 
     /// The parent item. If [Self::item] is e.g. an enum variant, then the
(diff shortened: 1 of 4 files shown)


Enselic:
Good news! We don't actually depend on https://github.com/aDotInTheVoid/rustdoc-types/pull/11 because I ended up not using the new `Module::is_stripped` field. My first solution made use of it, but not any more.

So I will merge this more or less right away. I am not comfortable to have nightly CI failing.

PR #78, opened 2022-07-17
IntermediatePublicItem: Remove `pub root: &'a Crate` field

Enselic:
I was always a bit sceptical of `IntermediatePublicItem::root`, but figured we might need it in the future, so I just let it be. I was expecting to need it when inlining imports, i.e. when doing https://github.com/Enselic/cargo-public-api/pull/77. But I didn't.

It was even in the way, because I needed to debug `IntermediatePublicItem`, but deriving `Debug` on it was not helpful, because it printed the entire crate (!).

So at this point it seems to create more problems than it solves. So I propose we remove it. If against all odds we need it in the future, we can add it back then.

Not only does that result in simpler code, it is also architecturally sound. It is better to not have each public item know about all other public items when it is not necessary.

And also  derive `Debug` on `IntermediatePublicItem` now.

### DIFF
--- a/public-api/src/intermediate_public_item.rs
+++ b/public-api/src/intermediate_public_item.rs
@@ -1,7 +1,7 @@
 use crate::render;
 use std::rc::Rc;
 
-use rustdoc_types::{Crate, Item};
+use rustdoc_types::Item;
 
 use crate::tokens::Token;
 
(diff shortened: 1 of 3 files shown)


douweschulte (approved):
I remember not liking the inclusion of it as well, but needing it for some information deep down. So I fully support removing it again.

Enselic:
Great! Thanks for review!

PR #80, opened 2022-07-21
Properly show re-exported primitive types

Enselic:
And also make comments more up to date and accurate.


Enselic:
Hiding from release notes because this works in the latest release, i.e. this PR fixes a regression in the unreleased change https://github.com/Enselic/cargo-public-api/pull/77

PR #94, opened 2022-08-12
Replace `parameterized_test!` macro with regular code

Enselic:
@douweschulte Are you OK with replacing `parameterized_test!` with regular code? I found the macro invocation hard to change, and so did my tooling (rust-analyzer + vscode).

I actually think the code becomes clearer without the macro. What do you think?

I needed to make adjustments to prepare for https://github.com/rust-lang/rust/pull/100335 (these adjustments will be made in a separate PR and is not part of this PR. Edit: See https://github.com/Enselic/cargo-public-api/pull/95 for the other PR which depends on this PR)

### DIFF
--- a/public-api/src/render.rs
+++ b/public-api/src/render.rs
@@ -790,10 +790,6 @@ mod test {
     use super::*;
     use rustdoc_types::Id;
 
-    // Tests for the `render_type` function.
-    // Missing:
-    //  * ImplTrait
-    //  * FunctionPointer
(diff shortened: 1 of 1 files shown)


douweschulte (approved):
Always change code that does not work anymore, so you have my full support. I am not sure if the expanded view is more clear or not. But I would assume it is easier for other collaborators, and that is a nice win in and on itself.

PR #107, opened 2022-08-21
Add build `--target` option

### LINKED ISSUE #105: Be able build for a specific target triple with `--target`
We have a set of crates with public APIs that we build and use only for a specific target, `wasm32-unknown-unknown`. Building with just `cargo public-api` results in no relevant API, as they require to be built with Cargo/rustdoc `--target wasm32-unknown-unknown`, so would be great to support such an option similar to how Cargo & rustdoc does itself.

Took a look at it and looks to be quite simple so may attempt it myself.

Related: 
- #100

repi:
This makes it possible to generate public API listing for a crate for a specific Rust build target other than the default.

Such as `--target x86_64-unknown-linux-musl` or `--target wasm32-unknown-unknown`

Resolves: #105

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -112,6 +112,10 @@ pub struct Args {
     /// Useful if you have built a toolchain from source for example.
     #[clap(long, hide = true, default_value = "+nightly")]
     rustdoc_json_toolchain: String,
+
+    /// Build for the target triple
+    #[clap(long)]
+    target: Option<String>,
 }
 
 /// After listing or diffing, we might want to do some extra work. This struct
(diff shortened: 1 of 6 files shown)


repi:
Would be good to have a test for this also, though that would be a bit more involved with having to install another target in CI and build and verify in the scripts. Less familiar with that part of this project.

Enselic:
I did a smartphone preliminary CR and it looks good, thanks! But yes we will need some kind of test. Can't we test with `--target x86_64-unknown-linux-gnu`? Basically to make sure the arg can be used and the JSON path becomes correct. With some luck you can copy-paste the `fn custom_toolchain()` test and change to use --target instead. I agree installing another target in CI is too complicated, so let's try to use the target that is already available  🙂

repi:
Ah yes one could actually just test with the same target that is the default, to at least verify that parameter is passed down and output read from the same path. That should work.

Ideally was imaging a test that has some types or fields in with `#[cfg(target_arch = "wasm32")]` so one in the test can verify that the output is actually different and that that type only shows up when building for the specific target. But having some basic validation without that sounds like a good first step.

Enselic:
Good point. Having `#[cfg(target_arch = "wasm32")]` or similar would be good. But I would be fine with just having the basic test discussed above before merging this initial version.

Enselic:
@repi I think I managed to write an OK test of this feature. It fails on main but passes on your branch. It looks like this: https://github.com/Enselic/cargo-public-api/commit/84b578817be4f585a85d7250e66fb04e893ac3a8

It is OK if I push that to this PR? I am asking, because I don't want to mess up any local changes you might have.

But if I may push that test and also merge origin/main to this PR, I think we can merge this PR afterwards if it passes CI.

repi:
looks good, and yeah please do push to the branch, think you have access to do so (but if not ping me here and I'll investigate).

definitely nice to get this merged in with the other recent fixes!

Enselic:
@repi Seems like I didn't have push rights, but you can do the push too. Should be as simple as this:
```
git fetch Enselic && git push EmbarkStudios Enselic/add-target-option:add-target-option
```
but please of course adjust to match your remote names.

repi:
pushed your additions to here now

Enselic:
Thank you! Let's merge this now then

PR #74, opened 2022-07-12
Make us work with recently released `rustup 1.25.0`

Enselic:
From `rustup 1.25.0` the env vars `RUSTDOC` and `RUSTC` are set to avoid having to resolve proxy settings more than once. Since we run under the stable toolchain, it means they  will point to stable tool chain tools. This overrides the toolchain we specify with `+`. Clear `RUSTENV` and `RUSTC` so we can run the `+nightly` toolchain.

See https://github.com/rust-lang/rustup/issues/3031#issuecomment-1181420211 (which talks about `RUSTC` but I confirmed the reasoning also applies to `RUSTDOC`).

Closes #73

PR #113, opened 2022-08-25
Make `cargo +custom public-api` build rustdoc JSON with toolchain `custom`

Emilgardis:
When using rustup shims, rustup sets and uses the `RUSTUP_TOOLCHAIN` variable which rustup and its shims use to get the specified or defaulted `+toolchain`.

With this change, `cargo +nightly public-api` or `cargo +stage2 public-api` etc. will work as expected.

This pr deprecates the `--toolchain` command as it's no longer needed.

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -159,7 +159,7 @@ fn main_() -> Result<()> {
     let mut args = get_args();
 
     // check if using a stable compiler, and use nightly if it is.
-    if probably_stable() {
+    if active_toolchain_is_probably_stable() {
         args.toolchain = Some("+nightly".to_owned());
     }
 
@@ -174,7 +174,12 @@ fn main_() -> Result<()> {
     post_processing.perform(&args)
 }
 
-fn probably_stable() -> bool {
+/// Returns true if it seems like the currently active toolchain is the stable
+/// toolchain.
+///
+/// See <https://rust-lang.github.io/rustup/overrides.html> for some
+/// more info of how different toolchains can be activated.
+fn active_toolchain_is_probably_stable() -> bool {
     let mut cmd = std::process::Command::new("cargo");
     cmd.arg("--version");
 
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -113,10 +113,13 @@ pub struct Args {
     verbose: bool,
 
     /// Allows you to build rustdoc JSON with a toolchain other than `+nightly`.
+    ///
+    /// Consider using `cargo +toolchain public-api` instead.
+    ///
     /// Useful if you have built a toolchain from source for example, or if you
     /// want to use a fixed toolchain in CI.
-    #[clap(long, default_value = "+nightly")]
-    toolchain: String,
+    #[clap(long)]
+    toolchain: Option<String>,
 
     /// Build for the target triple
     #[clap(long)]
@@ -153,7 +156,12 @@ struct PostProcessing {
 }
 
 fn main_() -> Result<()> {
-    let args = get_args();
+    let mut args = get_args();
+
+    // check if using a stable compiler, and use nightly if it is.
+    if probably_stable() {
+        args.toolchain = Some("+nightly".to_owned());
+    }
 
     let post_processing = if let Some(commits) = &args.diff_git_checkouts {
         print_diff_between_two_commits(&args, commits)?
@@ -164,6 +172,23 @@ fn main_() -> Result<()> {
     };
 
     post_processing.perform(&args)
+}
+
+fn probably_stable() -> bool {
+    let mut cmd = std::process::Command::new("cargo");
+    cmd.arg("--version");
+
+    let output = match cmd.output() {
+        Ok(output) => output,
+        Err(_) => return false,
+    };
+
+    let version = match String::from_utf8(output.stdout) {
+        Ok(version) => version,
+        Err(_) => return false,
+    };
+
+    version.starts_with("cargo 1") && !version.contains("nightly")
 }
 
 fn check_diff(args: &Args, diff: &Option<PublicItemsDiff>) -> Result<()> {
@@ -199,7 +224,7 @@ fn check_diff(args: &Args, diff: &Option<PublicItemsDiff>) -> Result<()> {
 }
 
 fn print_public_items_of_current_commit(args: &Args) -> Result<PostProcessing> {
-    let (public_items, branch_to_restore) = collect_public_api_from_commit(None)?;
+    let (public_items, branch_to_restore) = collect_public_api_from_commit(args, None)?;
 
     if args.verbose {
         public_items.missing_item_ids.iter().for_each(|i| {
@@ -219,10 +244,10 @@ fn print_public_items_of_current_commit(args: &Args) -> Result<PostProcessing> {
 
 fn print_diff_between_two_commits(args: &Args, commits: &[String]) -> Result<PostProcessing> {
     let old_commit = commits.get(0).expect("clap makes sure first commit exist");
-    let (old, branch_to_restore) = collect_public_api_from_commit(Some(old_commit))?;
+    let (old, branch_to_restore) = collect_public_api_from_commit(args, Some(old_commit))?;
 
     let new_commit = commits.get(1).expect("clap makes sure second commit exist");
-    let (new, _) = collect_public_api_from_commit(Some(new_commit))?;
+    let (new, _) = collect_public_api_from_commit(args, Some(new_commit))?;
 
     let diff_to_check = Some(print_diff(args, old.items, new.items)?);
 
@@ -301,9 +326,10 @@ fn get_options(args: &Args) -> Options {
 /// Collects public items from either the current commit or a given commit. If
 /// `commit` is `Some` and thus a `git checkout` will be made, also return the
 /// original branch.
-fn collect_public_api_from_commit(commit: Option<&str>) -> Result<(PublicApi, Option<String>)> {
-    let args = get_args();
-
+fn collect_public_api_from_commit(
+    args: &Args,
+    commit: Option<&str>,
+) -> Result<(PublicApi, Option<String>)> {
     // Do a git checkout of a specific commit unless we are supposed to simply
     // use the current commit
     let original_branch = if let Some(commit) = commit {
@@ -315,9 +341,8 @@ fn collect_public_api_from_commit(commit: Option<&str>) -> Result<(PublicApi, Op
     } else {
         None
     };
-
     let mut build_options = BuildOptions::default()
-        .toolchain(&args.toolchain)
+        .toolchain(args.toolchain.clone())
         .manifest_path(&args.manifest_path)
         .all_features(args.all_features)
         .no_default_features(args.no_default_features)
@@ -337,7 +362,7 @@ fn collect_public_api_from_commit(commit: Option<&str>) -> Result<(PublicApi, Op
     if args.verbose {
         println!("Processing {:?}", json_path);
     }
-    let options = get_options(&args);
+    let options = get_options(args);
 
     Ok((
         public_api_from_rustdoc_json_path(json_path, options)?,
--- a/public-api/tests/utils/mod.rs
+++ b/public-api/tests/utils/mod.rs
@@ -10,7 +10,7 @@ pub fn rustdoc_json_path_for_crate(test_crate: &str) -> PathBuf {
     // so build quietly to make running tests much less noisy
     rustdoc_json::build(
         BuildOptions::default()
-            .toolchain("+nightly")
+            .toolchain("+nightly".to_owned())
             .manifest_path(&format!("{}/Cargo.toml", test_crate))
             .quiet(true),
     )
--- a/rustdoc-json/src/build.rs
+++ b/rustdoc-json/src/build.rs
@@ -144,7 +144,15 @@ impl BuildOptions {
     /// Set the toolchain. Default: `None`.
     /// Until rustdoc JSON has stabilized, you will want to set this to
     /// be `"+nightly"` or similar.
+    ///
     /// If the toolchain is set as `None`, the current active toolchain will be used.
+    ///
+    /// # Notes
+    ///
+    /// The currently active toolchain is typically specified by the
+    /// `RUSTUP_TOOLCHAIN` environment variable, which the rustup proxy
+    /// mechanism sets. See <https://rust-lang.github.io/rustup/overrides.html>
+    /// for more info on how the active toolchain is determined.
     #[must_use]
     pub fn toolchain(mut self, toolchain: impl Into<Option<String>>) -> Self {
         self.toolchain = toolchain.into();
--- a/rustdoc-json/src/build.rs
+++ b/rustdoc-json/src/build.rs
@@ -2,7 +2,6 @@ use super::BuildError;
 use super::BuildOptions;
 
 use std::{
-    ffi::OsStr,
     path::{Path, PathBuf},
     process::Command,
 };
@@ -24,7 +23,7 @@ pub(crate) fn run_cargo_rustdoc(options: BuildOptions) -> Result<PathBuf, BuildE
         )
     } else {
         let manifest = cargo_toml::Manifest::from_path(&options.manifest_path)?;
-        if manifest.workspace.is_some() {
+        if manifest.package.is_none() && manifest.workspace.is_some() {
             Err(BuildError::VirtualManifest(options.manifest_path))
         } else {
             Err(BuildError::General(String::from("See above")))
@@ -50,12 +49,7 @@ fn cargo_rustdoc_command(options: &BuildOptions) -> Command {
     } = options;
     let mut command = Command::new("cargo");
 
-    // These can override our `+nightly` with `+stable` unless we clear them
-    command.env_remove("RUSTDOC");
-    command.env_remove("RUSTC");
-
-    let overridden_toolchain = OVERRIDDEN_TOOLCHAIN.map(OsStr::new);
-    if let Some(toolchain) = overridden_toolchain.or(requested_toolchain.as_deref()) {
+    if let Some(toolchain) = OVERRIDDEN_TOOLCHAIN.or(requested_toolchain.as_deref()) {
         command.arg(toolchain);
     }
 
@@ -147,12 +141,13 @@ impl Default for BuildOptions {
 }
 
 impl BuildOptions {
-    /// Set the toolchain. Default: `None`, which in practice means `"+stable"`.
+    /// Set the toolchain. Default: `None`.
     /// Until rustdoc JSON has stabilized, you will want to set this to
-    /// `"+nightly"` or similar.
+    /// be `"+nightly"` or similar.
+    /// If the toolchain is set as `None`, the current active toolchain will be used.
     #[must_use]
-    pub fn toolchain(mut self, toolchain: impl AsRef<OsStr>) -> Self {
-        self.toolchain = Some(toolchain.as_ref().to_owned());
+    pub fn toolchain(mut self, toolchain: impl Into<Option<String>>) -> Self {
+        self.toolchain = toolchain.into();
         self
     }
 
--- a/rustdoc-json/src/lib.rs
+++ b/rustdoc-json/src/lib.rs
@@ -8,7 +8,7 @@
 //!
 //!    let json_path = rustdoc_json::build(
 //!        BuildOptions::default()
-//!            .toolchain("+nightly")
+//!            .toolchain("+nightly".to_owned())
 //!            .manifest_path("Cargo.toml"),
 //!    ).unwrap();
 //!
@@ -21,7 +21,7 @@
 // deny in CI, only warn here
 #![warn(clippy::all, clippy::pedantic, missing_docs)]
 
-use std::{ffi::OsString, path::PathBuf};
+use std::path::PathBuf;
 
 mod build;
 
@@ -55,7 +55,7 @@ pub enum BuildError {
 /// See [crate] for an example on how to use it.
 #[derive(Debug)]
 pub struct BuildOptions {
-    toolchain: Option<OsString>,
+    toolchain: Option<String>,
     manifest_path: std::path::PathBuf,
     target: Option<String>,
     quiet: bool,
--- a/scripts/test-invocation-variants.sh
+++ b/scripts/test-invocation-variants.sh
@@ -77,5 +77,20 @@ assert_progress_and_output "cargo public-api"
 assert_progress_and_output "cargo public-api --manifest-path $(pwd)/Cargo.toml"
 
 # Make sure we can run the tool with MINIMUM_RUSTDOC_JSON_VERSION
-rustup toolchain install nightly-2022-08-15
+rustup toolchain install --no-self-update nightly-2022-08-15
 assert_progress_and_output "cargo +nightly-2022-08-15 public-api --manifest-path $(pwd)/Cargo.toml"
+
+# Sanity check to make sure we can make the tool build rustdoc JSON with a
+# custom toolchain via the rustup proxy mechanism (see
+# https://rust-lang.github.io/rustup/concepts/index.html#how-rustup-works). The
+# test uses a too old nightly toolchain, which should make the tool fail if it's used.
+custom_toolchain="nightly-2022-06-01"
+rustup toolchain install --no-self-update "${custom_toolchain}"
+cmd="cargo +${custom_toolchain} public-api"
+echo -n "${cmd} ... "
+if ${cmd} >/dev/null 2>/dev/null; then
+    echo "FAIL: Using '${custom_toolchain}' to build rustdoc JSON should have failed!"
+    exit 1
+else
+    echo "PASS"
+fi
--- a/scripts/test-invocation-variants.sh
+++ b/scripts/test-invocation-variants.sh
@@ -75,3 +75,7 @@ assert_progress_and_output "cargo public-api"
 
 # Make sure we can run the tool on an external directory as a cargo sub-command
 assert_progress_and_output "cargo public-api --manifest-path $(pwd)/Cargo.toml"
+
+# Make sure we can run the tool with MINIMUM_RUSTDOC_JSON_VERSION
+rustup toolchain install nightly-2022-08-15
+assert_progress_and_output "cargo +nightly-2022-08-15 public-api --manifest-path $(pwd)/Cargo.toml"
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -61,7 +61,7 @@ You can also combine both ways:
 If you have built rustdoc yourself to try some rustdoc JSON fix, you can run `cargo public-api` with your [custom toolchain](https://rustc-dev-guide.rust-lang.org/building/how-to-build-and-run.html#creating-a-rustup-toolchain) like this:
 
 ```
-cargo public-api --toolchain +custom
+cargo +custom public-api
 ```
 
 Another option is the `RUSTDOC_JSON_OVERRIDDEN_TOOLCHAIN_HACK` env var. Use it like this:
--- a/rustdoc-json/examples/build-rustdoc-json.rs
+++ b/rustdoc-json/examples/build-rustdoc-json.rs
@@ -8,7 +8,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Build it
     let json_path = rustdoc_json::build(
         BuildOptions::default()
-            .toolchain("+nightly")
+            .toolchain("+nightly".to_owned())
             .manifest_path(&std::env::args().nth(1).unwrap()),
     )?;
     println!("Built and wrote rustdoc JSON to {:?}", &json_path);
--- a/rustdoc-json/public-api.txt
+++ b/rustdoc-json/public-api.txt
@@ -19,7 +19,7 @@ pub fn rustdoc_json::BuildOptions::no_default_features(self, no_default_features
 pub fn rustdoc_json::BuildOptions::package(self, package: impl AsRef<str>) -> Self
 pub fn rustdoc_json::BuildOptions::quiet(self, quiet: bool) -> Self
 pub fn rustdoc_json::BuildOptions::target(self, target: String) -> Self
-pub fn rustdoc_json::BuildOptions::toolchain(self, toolchain: impl AsRef<OsStr>) -> Self
+pub fn rustdoc_json::BuildOptions::toolchain(self, toolchain: impl Into<Option<String>>) -> Self
 pub fn rustdoc_json::build(options: BuildOptions) -> Result<PathBuf, BuildError>
 pub mod rustdoc_json
 pub struct rustdoc_json::BuildOptions


rustdoc-json/src/build.rs
```diff
@@ -26,7 +27,9 @@ pub(crate) fn run_cargo_rustdoc(options: BuildOptions) -> Result<PathBuf, BuildE
         rustdoc_json_path_for_manifest_path(options.manifest_path, options.target.as_deref())
     } else {
         let manifest = cargo_toml::Manifest::from_path(&options.manifest_path)?;
-        if manifest.workspace.is_some() {
+        if manifest.package.is_some() {
+            Err(BuildError::RustDocError)
+        } else if manifest.workspace.is_some() {
             Err(BuildError::VirtualManifest(options.manifest_path))
         } else {
             Err(BuildError::General(String::from("See above")))
```
Emilgardis: Maybe this should actually be

```suggestion
        if manifest.package.is_none() && manifest.workspace.is_some() {
            Err(BuildError::VirtualManifest(options.manifest_path))
        } else {
            Err(BuildError::General(String::from("See above")))
```

Emilgardis: fixed

cargo-public-api/src/main.rs
```diff
@@ -108,11 +108,16 @@ pub struct Args {
     #[clap(long, hide = true)]
     verbose: bool,
 
+    /// [DEPRECATED]
```
Enselic: --toolchain has not been released yet so let's just outright remove this arg on cargo-public-api instead of deprecating it

rustdoc-json/src/lib.rs
```diff
@@ -52,6 +53,7 @@ pub enum BuildError {
 /// See [crate] for an example on how to use it.
 #[derive(Debug)]
 pub struct BuildOptions {
+    #[deprecated(note = "use `RUSTUP_TOOLCHAIN` or `cargo +toolchain` instead")]
     toolchain: Option<OsString>,
```
Enselic: Hmm, I'm not sure I'm a fan of a library reading "magical" environment vars behind the scenes. What do you think about introducing an enum instead? Pseudo code:

```rust
enum Toolchain {
    FromRustupToolchainEnvVar,
    Named(OsString),
}
```

and then keep

```rust
    toolchain: Option<Toolchain>,
```

without deprecation.

that way the API surface and library behaviour and becomes clearer IMHO.

rustdoc-json/src/build.rs
```diff
@@ -47,13 +48,12 @@ fn cargo_rustdoc_command(
...
     let mut command = Command::new("cargo");
 
-    // These can override our `+nightly` with `+stable` unless we clear them
-    command.env_remove("RUSTDOC");
-    command.env_remove("RUSTC");
-
     let overridden_toolchain = OVERRIDDEN_TOOLCHAIN.map(OsStr::new);
     if let Some(toolchain) = overridden_toolchain.or(requested_toolchain) {
         command.arg(toolchain);
+        // These can override our `+nightly` with `+stable` unless we clear them
```
Enselic: I added these lines as a fix for https://github.com/rust-lang/rustup/issues/3031#issuecomment-1181420211, see https://github.com/Enselic/cargo-public-api/pull/74

However, after just one day or so the rustup folks fixed this regression and released a patch release, so the fix became obsoleted. I kept it around in case the regression would show up again.

But now I think we should just remove these three lines rather than trying to make it work with the new way to pick toolchain.

Unless you know that they are still needed?

rustdoc-json/src/lib.rs
```diff
@@ -6,9 +6,10 @@
 //! ```
 //!    use rustdoc_json::BuildOptions;
 //!
+//!     std::env::set_var("RUSTUP_TOOLCHAIN", "nightly");
+//!
 //!    let json_path = rustdoc_json::build(
 //!        BuildOptions::default()
-//!            .toolchain("+nightly")
```
Enselic, line 11: I think we want this to be something like `.toolchain(Toolchain::Named("+nightly"))` instead (see my other comment)

rustdoc-json/src/build.rs
```diff
@@ -26,7 +27,7 @@ pub(crate) fn run_cargo_rustdoc(options: BuildOptions) -> Result<PathBuf, BuildE
         rustdoc_json_path_for_manifest_path(options.manifest_path, options.target.as_deref())
     } else {
         let manifest = cargo_toml::Manifest::from_path(&options.manifest_path)?;
-        if manifest.workspace.is_some() {
+        if manifest.package.is_none() && manifest.workspace.is_some() {
```
Enselic, line 26: is this change related to removal of --toolchain? or is it some other fix?

.github/workflows/CI.yml
```diff
@@ -59,6 +60,7 @@ jobs:
         with:
           toolchain: nightly
           profile: minimal
+          override: true
```
Enselic: How come this is needed? I'd like CI to be as similar as possible to real user machines, and they will typically not have this override

Maybe it is not needed if we add an enum as suggested in another comment?

Enselic (changes requested):
Thank you so much. Looking forward to collaborate with you to get this merged. I think this change makes sense.

At first I wasn't sure if the `+` syntax was canonical, since typically `+` means a different binary will be used (e.g. nightly rustdoc vs stable rustdoc), whereas in our case the same top-level binary is always used: "cargo-public-api". But I think this makes enough sense to go ahead with, because INDRECTLY a different `rustdoc` binary is used.

Another inconsistency and potential source of confusion: omitting `+` normally means "use stable", which is not the case for us. But again, I think we can get away with that.

I added some comments. I am by no means all-knowledgeble, so if you think I am on the wrong track on some or all of my comments, please speak up.

Thanks again! Looking forward to getting this merged after we have sorted out the open questions!

cargo-public-api/src/main.rs
Emilgardis: awesome

rustdoc-json/src/lib.rs
Emilgardis: Makes sense to me! I think removing the toolchain flag from public api, but keeping this is good enough

rustdoc-json/src/build.rs
Emilgardis: I wasn't sure why they were needed and kept them just because they were there, removing sounds good

rustdoc-json/src/build.rs
Emilgardis: It's a fix, a virtual manifest is one without a package defined. For example you can have the following `Cargo.toml`

```toml
[package]
name = "my_crate"
...

[workspace]
members = ["my_other_crate"]
```

see https://github.com/cross-rs/test-workspace/blob/main/workspace/Cargo.toml for example which has all possible alternatives (except a virtual manifest, which would be a Cargo.toml without `package`

.github/workflows/CI.yml
Emilgardis: specifying `override` makes it so the toolchain specified with `actions-rs/toolchain` gets set as the one to be used by subsequent `cargo`/`rustc`/`rustdoc` etc, it's equivalent to having a `rust-toolchain.toml` file or running `rustup override set <toolchain>`

Emilgardis: Basically, it's to not have to do `cargo +nightly <command>` in ci, because public-api doesn't set a default now, but uses the current set rustup toolchain. This could possible be changed similarly to how `cargo-expand` does it, specify nightly if the current toolchain is definitely not nightly. https://github.com/dtolnay/cargo-expand/blob/b8b0a9d658430b51e6142c65222cd06249b39506/src/main.rs#L57.

I can implement this behaviour, I think I actually prefer it.

rustdoc-json/src/build.rs
Enselic: Thanks! 

Since this code will be hit only if building rustdoc fails, and since it typically won't fail if both package and workspace is present, I think this is edge case enough to not mention in release notes and thus can piggy-back on this PR. (If we want to mention this fix in the release notes we need to split it out to a different PR since I auto-generate release notes from PR titles.)

Emilgardis:
will squash if all looks good

one comment from me as a contributor, I'd hugely recommend adding a `#[warn(...)]` for everything that is gated by ci, confusing with no clippy warns locally (and not using the script)

cargo-public-api/src/main.rs
```diff
@@ -135,6 +129,11 @@ struct PostProcessing {
 fn main_() -> Result<()> {
     let args = get_args();
 
+    // check if using a stable compiler, and if not use nightly.
+    if probably_stable() && std::env::var("CARGO_PUBLIC_API_NO_RUN_NIGHTLY").is_err() {
+        std::env::set_var("RUSTUP_TOOLCHAIN", "nightly");
```
Enselic: Using `cargo-expand` as a role model makes a lot of sense, but I am a bit confused, because when I git grep in `cargo-expand` it does not seem to read `RUSTUP_TOOLCHAIN`. Do you know if it uses some helper crate for that perhaps?

.github/workflows/CI.yml
```diff
@@ -48,6 +48,7 @@ jobs:
...
```
Enselic: You forgot to remove this it seems?

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -34,13 +34,6 @@ fn list_public_items_with_lint_error() {
         .success();
 }
 
-#[test]
-fn custom_toolchain() {
```
Enselic: I'd like us to keep this test since it is important to make sure that `cargo public-api` can use a custom toolchain. Couldn't we just change invocation here to match? If it becomes a non-sensical test to test with `+nightly` I think we should also install `nightly-2022-08-15` (current `MINIMUM_RUSTDOC_JSON_VERSION`) so we can pass that instead.

Enselic (commented):
Thank you for the update! I did another review round and the new enum look good to me. Had some  more questions/comments though.

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Emilgardis: what about

```
#[test]
#[ignore]
fn minimum_toolchain() {
	// set BuildOptions.toolchain to `nightly-2022-08-15`, do the other stuff.
}
```

Adding a CI test for this makes sense

.github/workflows/CI.yml
Emilgardis: It needs to be left in as (some of) the tests assume that the toolchain is changed in the code, but the toolchain is only changed in `main_` and not in the tests

cargo-public-api/src/main.rs
Emilgardis: `cargo-expand` reruns itself but invoked as `rustup run nightly cargo expand` (or simply put `cargo +nightly expand`. It's not strictly needed, and setting `RUSTUP_TOOLCHAIN=nightly` achieves the same result once a rustup shim is called.

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Emilgardis: I opted with adding this to test-invocation-variants.sh

cargo-public-api/src/main.rs
Enselic: I did a bit of a rushed review this morning since I wanted to give you some feedback before taking kids to school and going to my day job. I've done some more thinking during the day, and here is where my thoughts currently are:

My initial idea with `FromRustupToolchainEnvVar` was to let the library read `RUSTUP_TOOLCHAIN`. Shouldn't that work? I think that will allow us to get rid of the `override` in CI. In fact, the current enum is only semantically different over `Option<String>`, isn't it? `Default` = `None`, `Named(foo)` = `Some(foo)`.

Also, what is `CARGO_PUBLIC_API_NO_RUN_NIGHTLY` about?

Again, I really appreciate you working on this, and I hope you don't mind us discussing this a bit back and forth.

I'll wait with another detailed review round until we have settled these higher-level questions.

I did see your other PRs 🎉 I will take a look right away!

Emilgardis: Why would the library read `RUSTUP_TOOLCHAIN` ? It's already set in env, and doing `cargo +$RUSTUP_TOOLCHAIN <cmd>` just does the same thing as `cargo <cmd>`

`CARGO_PUBLIC_API_NO_RUN_NIGHTLY` is for not setting the toolchain to nightly if the current toolchain is stable, this is to make it possible to use a stable compiler with the binary. Technically not needed, as `-Z unstable-options` will always need a unstable toolchain

Emilgardis: > Shouldn't that work? I think that will allow us to get rid of the override in CI

Try running `cargo test` with a stable toolchain and you'll see the issue.

```
❯ cargo test
   Compiling rustdoc-json v0.3.1 (/Users/emil/workspace/cargo-public-api/rustdoc-json)
   Compiling cargo-public-api v0.14.0 (/Users/emil/workspace/cargo-public-api/cargo-public-api)
   Compiling public-api v0.15.0 (/Users/emil/workspace/cargo-public-api/public-api)
    Finished test [unoptimized + debuginfo] target(s) in 2.51s
     Running unittests src/lib.rs (target/debug/deps/cargo_public_api-cad645bec020de87)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/main.rs (target/debug/deps/cargo_public_api-fdca03bee1ae1ceb)

running 3 tests
test arg_types::tests::test_deny_added ... ok
test arg_types::tests::test_deny_changed ... ok
test arg_types::tests::test_deny_removed ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/cargo-public-api-bin-tests.rs (target/debug/deps/cargo_public_api_bin_tests-9b155849dae31082)

running 31 tests
test deny_with_invalid_arg ... ok
test deny_added_when_not_diffing ... ok
test deny_when_not_diffing ... ok
test deny_removed_when_not_diffing ... ok
test deny_changed_when_not_diffing ... ok
test deny_combination_when_not_diffing ... ok
test diff_public_items_from_files ... ok
test diff_public_items_missing_one_arg ... ok
test deny_removed_with_diff ... ok
test deny_added_with_diff ... ok
test deny_without_diff ... ok
test deny_changed_with_diff ... ok
test deny_with_diff ... ok
test diff_public_items_without_git_root ... ok
test diff_public_items_with_dirty_tree_fails ... ok
test diff_public_items_markdown ... ok
test diff_public_items ... ok
test long_help ... ok
test short_help ... ok
test diff_public_items_markdown_no_changes ... ok
test diff_public_items_detached_head ... ok
test virtual_manifest_error ... ok
test diff_public_items_with_color ... ok
test list_public_items ... ok
test diff_public_items_with_manifest_path ... ok
test list_public_items_with_lint_error ... ok
test list_public_items_markdown ... ok
test list_public_items_explicit_manifest_path ... ok
test target_arg ... ok
test list_public_items_with_color ... ok
test verbose ... ok

test result: ok. 31 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.62s

     Running tests/cargo-public-api-readme-tests.rs (target/debug/deps/cargo_public_api_readme_tests-3dafe32ee512dee9)

running 1 test
test installation_instructions_mentions_minimum_rustdoc_json_version ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/public_api-bda5eb184e903f0e)

running 25 tests
test diff::tests::no_diff_means_empty_diff ... ok
test diff::tests::single_and_only_item_removed ... ok
test diff::tests::middle_item_added ... ok
test diff::tests::middle_item_removed ... ok
test diff::tests::single_and_only_item_added ... ok
test render::test::test_type_generic ... ok
test render::test::test_type_infer ... ok
test render::test::test_type_array ... ok
test render::test::test_type_path ... ok
test diff::tests::no_off_by_one_diff_skewing ... ok
test render::test::test_type_pointer ... ok
test render::test::test_type_pointer_mut ... ok
test render::test::test_type_primitive ... ok
test diff::tests::many_identical_items ... ok
test render::test::test_type_ref ... ok
test render::test::test_type_ref_lt ... ok
test render::test::test_type_ref_lt_mut ... ok
test render::test::test_type_ref_mut ... ok
test render::test::test_type_resolved_crate_name ... ok
test render::test::test_type_resolved_long_name ... ok
test render::test::test_type_resolved_simple ... ok
test render::test::test_type_slice ... ok
test render::test::test_type_resolved_name_crate ... ok
test render::test::test_type_tuple_empty ... ok
test render::test::test_type_tuple ... ok

test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/main.rs (target/debug/deps/public_api-67745131aabc0095)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/public-api-bin-tests.rs (target/debug/deps/public_api_bin_tests-7c1cbce8c2f73565)

running 10 tests
error: the option `Z` is only accepted on the nightly compiler
error: the option `Z` is only accepted on the nightly compiler


error: could not document `example_api`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2021 --crate-type lib --crate-name example_api src/lib.rs -o /Users/emil/workspace/cargo-public-api/test-apis/example_api-v0.1.0/target/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat -Z unstable-options --output-format json --cap-lints warn -C metadata=e7754d7af416a103 -L dependency=/Users/emil/workspace/cargo-public-api/test-apis/example_api-v0.1.0/target/debug/deps --crate-version 0.1.0` (exit status: 1)
error: could not document `example_api`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2021 --crate-type lib --crate-name example_api src/lib.rs -o /Users/emil/workspace/cargo-public-api/test-apis/example_api-v0.2.0/target/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat -Z unstable-options --output-format json --cap-lints warn -C metadata=f9e99507877cdd94 -L dependency=/Users/emil/workspace/cargo-public-api/test-apis/example_api-v0.2.0/target/debug/deps --crate-version 0.2.0` (exit status: 1)
test print_diff ... FAILED
test print_public_api_with_blanket_implementations ... FAILED
test no_args_shows_help ... ok
test long_help ... ok
test too_many_args_shows_help ... ok
test short_help ... ok
error: the option `Z` is only accepted on the nightly compiler

error: could not document `example_api`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2021 --crate-type lib --crate-name example_api src/lib.rs -o /Users/emil/workspace/cargo-public-api/test-apis/example_api-v0.2.0/target/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat -Z unstable-options --output-format json --cap-lints warn -C metadata=f9e99507877cdd94 -L dependency=/Users/emil/workspace/cargo-public-api/test-apis/example_api-v0.2.0/target/debug/deps --crate-version 0.2.0` (exit status: 1)
test print_diff_reversed ... FAILED
error: the option `Z` is only accepted on the nightly compiler

error: could not document `comprehensive_api`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2021 --crate-type lib --crate-name comprehensive_api src/lib.rs -o /Users/emil/workspace/cargo-public-api/test-apis/comprehensive_api/target/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat -Z unstable-options --output-format json --cap-lints warn -C metadata=cc391f1f93a5e47e -L dependency=/Users/emil/workspace/cargo-public-api/test-apis/comprehensive_api/target/debug/deps --extern rand=/Users/emil/workspace/cargo-public-api/test-apis/comprehensive_api/target/debug/deps/librand-c26db38e8be2dce6.rmeta --crate-version 0.1.0` (exit status: 1)
test broken_pipe ... FAILED
error: the option `Z` is only accepted on the nightly compiler

error: could not document `example_api`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2021 --crate-type lib --crate-name example_api src/lib.rs -o /Users/emil/workspace/cargo-public-api/test-apis/example_api-v0.2.0/target/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat -Z unstable-options --output-format json --cap-lints warn -C metadata=f9e99507877cdd94 -L dependency=/Users/emil/workspace/cargo-public-api/test-apis/example_api-v0.2.0/target/debug/deps --crate-version 0.2.0` (exit status: 1)
test print_no_diff ... FAILED
error: the option `Z` is only accepted on the nightly compiler

error: could not document `comprehensive_api`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2021 --crate-type lib --crate-name comprehensive_api src/lib.rs -o /Users/emil/workspace/cargo-public-api/test-apis/comprehensive_api/target/doc --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat -Z unstable-options --output-format json --cap-lints warn -C metadata=cc391f1f93a5e47e -L dependency=/Users/emil/workspace/cargo-public-api/test-apis/comprehensive_api/target/debug/deps --extern rand=/Users/emil/workspace/cargo-public-api/test-apis/comprehensive_api/target/debug/deps/librand-c26db38e8be2dce6.rmeta --crate-version 0.1.0` (exit status: 1)
test print_public_api ... FAILED

failures:

---- print_diff stdout ----
thread 'print_diff' panicked at 'called `Result::unwrap()` on an `Err` value: General("See above")', public-api/tests/utils/mod.rs:16:6

---- print_public_api_with_blanket_implementations stdout ----
thread 'print_public_api_with_blanket_implementations' panicked at 'called `Result::unwrap()` on an `Err` value: General("See above")', public-api/tests/utils/mod.rs:16:6

---- print_diff_reversed stdout ----
thread 'print_diff_reversed' panicked at 'called `Result::unwrap()` on an `Err` value: General("See above")', public-api/tests/utils/mod.rs:16:6

---- broken_pipe stdout ----
thread 'broken_pipe' panicked at 'called `Result::unwrap()` on an `Err` value: General("See above")', public-api/tests/utils/mod.rs:16:6

---- print_no_diff stdout ----
thread 'print_no_diff' panicked at 'called `Result::unwrap()` on an `Err` value: General("See above")', public-api/tests/utils/mod.rs:16:6

---- print_public_api stdout ----
thread 'print_public_api' panicked at 'called `Result::unwrap()` on an `Err` value: General("See above")', public-api/tests/utils/mod.rs:16:6


failures:
    broken_pipe
    print_diff
    print_diff_reversed
    print_no_diff
    print_public_api
    print_public_api_with_blanket_implementations

test result: FAILED. 4 passed; 6 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.21s
```

I don't think it makes sense to run the tests using another toolchain than the current one.

Enselic:
> one comment from me as a contributor, I'd hugely recommend adding a #[warn(...)] for everything that is gated by ci, confusing with no clippy warns locally (and not using the script)

So far I have deliberately stayed away from enforcing lints in code, because that makes e.g. `git bisect` in a code base a nightmare. With `#[deny(...)]` at least. Although that can we worked around with `--cap-lints`. It _should_ be fine to have `#[warn(...)]` I guess.

I weight your input as a contributor very high, so I will add `#[warn(...)]` in the code in the near future,  and  keep `deny` in CI. I think the downside of duplicated configs and warnings during e.g. `git bisect` is smaller than the upside of having lint warnings "in-band". 

Feel free to create a PR for that too! (But you are absolutely not required to do so of course, I will do it myself as soon as I get time)

Emilgardis:
Yep, deny in code is generally bad, warn is fine though. See https://github.com/rust-unofficial/patterns/blob/main/anti_patterns/deny-warnings.md 

using deny can also make it so that consumers will fail if new lints are added and you deny on a group (like `clippy::pedantic)`

public-api/tests/utils/mod.rs
```diff
@@ -10,7 +10,6 @@ pub fn rustdoc_json_path_for_crate(test_crate: &str) -> PathBuf {
     // so build quietly to make running tests much less noisy
     rustdoc_json::build(
         BuildOptions::default()
-            .toolchain("+nightly")
```
Enselic, line 13: All tests pass for me locally if we keep this as `.toolchain(Toolchain::named("+nightly"))`. But I'm not sure we need to the new `Toolchain` enum at all any longer? (See other thread.)

.github/workflows/CI.yml
Enselic: Isn't that solved by not stopping to specify `+nightly`? See other comment

cargo-public-api/src/main.rs
Enselic: Ah, right, I didn't consider the effect of `RUSTUP_TOOLCHAIN` being inherited by the `cargo` process spawned from inside `rustdoc_json`. But in that case, why do we need to change the `rustdoc_json` API at all? I think it makes sense to still be able to manually specify a toolchain with the lib, because that has higher prio than `RUSTUP_TOOLCHAIN` when looking at https://rust-lang.github.io/rustup/overrides.html.

And if we do that, I wonder if it wouldn't be more elegant to pass `+nightly` as `toolchain`  from `cargo-public-api`  if `probably_stable()`, instead of setting `RUSTUP_TOOLCHAIN`?

And do we even need `probably_stable()` in its current "complicated" form? Couldn't we just check if `RUSTUP_TOOLCHAIN` is "stable" inside of it?

When I run
```
cargo +nightly run  -- --manifest-path cargo-public-api/Cargo.toml
```

I get:

```
RUSTUP_TOOLCHAIN=nightly-aarch64-apple-darwin
```

in the env, and with

```
cargo run  -- --manifest-path cargo-public-api/Cargo.toml
```

I get

```
RUSTUP_TOOLCHAIN=1.63-aarch64-apple-darwin
```

Seems like we can use that instead of running `cargo` inside of `probably_stable()`.

I added this in main.rs to inspect env:

```
    std::process::Command::new("env").spawn().unwrap().wait().unwrap();
```

Emilgardis: The problem with using `RUSTUP_TOOLCHAIN` is that it gives no real information about the version of the toolchain. `--version` or reading the `multirust-channel-manifest.toml` file is the only reliable way (and reading `multirust-channel-manifest.toml` only works on normal toolchains, not linked ones (`rustup toolchain link`))

public-api/tests/utils/mod.rs
Emilgardis: If I do this, how would one run the tests with `+nightly-2022-08-15` for example?

cargo-public-api/src/main.rs
Emilgardis: > I think it makes sense to still be able to manually specify a toolchain with the lib
> ...
> And if we do that, I wonder if it wouldn't be more elegant to pass +nightly as toolchain from cargo-public-api if probably_stable(), instead of setting RUSTUP_TOOLCHAIN?

I'd then have to reintroduce the `--toolchain` flag to be able to hand it down, or have it as a argument to some methods. Could also keep it but make it hidden (and unsettable)

also, the overrides doc is a bit misleading here. `+toolchain` only has one effect, setting the `RUSTUP_TOOLCHAIN`

Emilgardis: I've readded toolchain but made it skipped by clap

Emilgardis:
One thing I could do, is set the tests

```
broken_pipe
print_diff
print_diff_reversed
print_no_diff
print_public_api
print_public_api_with_blanket_implementations
```

behind a `#[cfg_attr(not(nightly), ignore)]` by introducing a build script which does

```rs

use rustc_version::{version, version_meta, Channel};

fn main() {
    // Assert we haven't travelled back in time
    assert!(version().unwrap().major >= 1);

    if matches!(version_meta().unwrap().channel, Channel::Nightly | Channel::Dev) {
        println!("cargo:rustc-cfg=nightly");
    }
}
```

Enselic:
It became too complex to continue with code review comments to communicate my vision for how to solve this problem, so I wrote some code instead: https://github.com/Enselic/cargo-public-api/pull/119

Looking forward to your input!

Emilgardis:
I've revamped this a bit, I now understand what you were going for with your comments after seeing #119 and I think this should satisfy that now

public-api/tests/utils/mod.rs
Emilgardis: Answered in https://github.com/Enselic/cargo-public-api/pull/119#issuecomment-1229257047

use `RUSTDOC_JSON_OVERRIDDEN_TOOLCHAIN_HACK`

i've readded this now

scripts/test-invocation-variants.sh
```diff
@@ -75,3 +75,7 @@ assert_progress_and_output "cargo public-api"
 
 # Make sure we can run the tool on an external directory as a cargo sub-command
 assert_progress_and_output "cargo public-api --manifest-path $(pwd)/Cargo.toml"
+
```
Enselic, line 78: This test also pass on origin/main. (I know it was my idea to use nightly-2022-08-15 . Sorry :) ).

I came up with this test:

```bash
# Sanity check to make sure we can make the tool build rustdoc JSON with a
# custom toolchain via the rustup proxy mechanism (see
# https://rust-lang.github.io/rustup/concepts/index.html#how-rustup-works). The
# test uses a too old nightly toolchain, which should make the tool fail if it
# works to use that toolchain
custom_toolchain="nightly-2022-06-01"
rustup toolchain install "${custom_toolchain}"
cmd="cargo +${custom_toolchain} public-api"
echo -n "${cmd} ... "
if ${cmd} >/dev/null 2>/dev/null; then
    echo "FAIL: Using '${custom_toolchain}' to build rustdoc JSON should have failed!"
    exit 1
else
    echo "PASS"
fi
```

Not very elegant, but it fails on main but passes with this PR at least, which is an important property.

cargo-public-api/src/main.rs
```diff
@@ -166,6 +174,23 @@ fn main_() -> Result<()> {
     post_processing.perform(&args)
 }
 
+fn probably_stable() -> bool {
```
Enselic: I like code to be close to overly documented and clear (for the benefit of new contributors), so may I propose we change to this please?

```suggestion
/// Returns true if it seems like the currently active toolchain is the stable
/// toolchain. See <https://rust-lang.github.io/rustup/overrides.html> for some
/// more info of how different toolchains can be activated.
fn active_toolchain_is_probably_stable() -> bool {
```

cargo-public-api/src/main.rs
```diff
@@ -301,9 +326,10 @@ fn get_options(args: &Args) -> Options {
 /// Collects public items from either the current commit or a given commit. If
 /// `commit` is `Some` and thus a `git checkout` will be made, also return the
 /// original branch.
-fn collect_public_api_from_commit(commit: Option<&str>) -> Result<(PublicApi, Option<String>)> {
-    let args = get_args();
```
Enselic, line 305: Thank you for getting around to get rid of the double invocation of `get_args()`. That has bugged me for a while, I just never got around fixing it.

rustdoc-json/src/lib.rs
```diff
@@ -8,7 +8,7 @@
 //!
 //!    let json_path = rustdoc_json::build(
 //!        BuildOptions::default()
-//!            .toolchain("+nightly")
+//!            .toolchain("+nightly".to_owned())
```
Enselic, line 11: Would be nice to somehow be able to keep the old code since it is simpler, but I couldn't figure how a good way to do that after trying quickly, so let's go with this.

rustdoc-json/src/build.rs
```diff
@@ -147,12 +141,13 @@ impl Default for BuildOptions {
 }
 
 impl BuildOptions {
-    /// Set the toolchain. Default: `None`, which in practice means `"+stable"`.
+    /// Set the toolchain. Default: `None`.
     /// Until rustdoc JSON has stabilized, you will want to set this to
-    /// `"+nightly"` or similar.
+    /// be `"+nightly"` or similar.
+    /// If the toolchain is set as `None`, the current active toolchain will be used.
```
Enselic, line 148: Can we amend the docs with this please?

```rust
    /// The currently active toolchain is typically specified by the
    /// `RUSTUP_TOOLCHAIN` environment variable, which the rustup proxy
    /// mechanism sets. See <https://rust-lang.github.io/rustup/overrides.html>
    /// for more info on how the active toolchain is determined.
```

As a side-note, if we were to make `rustdoc_json` use the rustc private crates directly, like `rustdoc` itself, we could make building rustdoc less magical, because it would not depend on rustup proxy stuff. The downside would be that it would require `rustdoc_json` to be regularly updated to match any change in rustc internals. That is an adventure for another time though.

Enselic (commented):
I did another review round. No big remarks. Thanks!

After comments have been addressed I think we can merge this, unless anyone of us thinks of more things to tweak.

scripts/test-invocation-variants.sh
```diff
@@ -50,28 +50,47 @@ assert_progress_and_output() {
     echo "PASS"
 }
 
-# Now we are ready to run the actual tests
+        # # Now we are ready to run the actual tests
```
Enselic: (You forgot to uncomment these)

Emilgardis: oops!

Enselic (approved):
Thank you! Let's merge!

I plan to make a release in the coming days. I think I'll do some minor cleanups first, like adding more elaborate --help texts to all new CLI args.

But it should not take many days for a release to happen.

Thank you again for your large batch of very useful PRs :)

Enselic:
Meh, who wants to wait for releases? Not me.

Released as https://github.com/Enselic/cargo-public-api/releases/tag/v0.15.0 (🥳)

PR #115, opened 2022-08-26
Support for `--features` / `--all-features` / `--no-default-features`

### LINKED ISSUE #100: Be able to enable `--features` / `--all-features` / `--no-default-features`
Would it be possible to add support for the Cargo `--features` / `--all-features` / `--no-default-features` options? Pipe them through to the underlying Cargo invocation.

We have multiple crates where we are both interested in the public API with `--all-features` enabled and also some where one wants to check for specific `--features`.

For example our [`puffin`](https://github.com/EmbarkStudios/puffin) crate has a mutually exclusive features (not ideal) and fails to analyse:

```
$ cargo public-api --manifest-path ./puffin/Cargo.toml
    Checking once_cell v1.9.0
    Checking byteorder v1.4.3
 Documenting puffin v0.12.1 (/Users/repi/git/embark/puffin/puffin)
error: Either feature zstd or ruzstd must be enabled
   --> puffin/src/frame_data.rs:491:1
    |
491 | compile_error!("Either feature zstd or ruzstd must be enabled");
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: Compilation failed, aborting rustdoc

error: could not document `puffin`

Caused by:
  process didn't exit successfully: `rustdoc --edition=2018 --crate-type lib --crate-name puffin puffin/src/lib.rs -o /Users/repi/git/embark/puffin/target/doc --cfg 'feature="default"' --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat -Z unstable-options --output-format json --cap-lints warn -C metadata=00aac93aa62b9ad4 -L dependency=/Users/repi/git/embark/puffin/target/debug/deps --extern byteorder=/Users/repi/git/embark/puffin/target/debug/deps/libbyteorder-922739cc82a0baa2.rmeta --extern once_cell=/Users/repi/git/embark/puffin/target/debug/deps/libonce_cell-b334803526b5c51d.rmeta --crate-version 0.12.1` (exit status: 1)
Error: Failed to build rustdoc JSON. Stderr: See above
```

Emilgardis:
enables `--features` / `--all-features` / `--no-default-features`

resolves #100


rustdoc-json/src/build.rs
```diff
@@ -43,6 +46,9 @@ fn cargo_rustdoc_command(
     requested_toolchain: Option<&OsStr>,
     manifest_path: impl AsRef<Path>,
     target: Option<&str>,
+    features: &[String],
+    no_default_features: bool,
+    all_features: bool,
     quiet: bool,
 ) -> Command {
```
Emilgardis: `cargo_rustdoc_command` should really take a `&BuildOptions`

Enselic (commented):
Looks great! I have no comments on the code (except nitpicks, that I will refrain from making). Thanks!

Now we only need a bunch of tests before we can merge this. I think adding some simple features to `test-apis/comprehensive_api` and writing some tests for that would make good tests. We need at least basic tests for all flags, and preferably some test that tests combination of flags. Doesn't have to be super exhaustive though.

If you are not excited about writing tests, let me know, and I will help out when I get time.

Emilgardis:
Added a test, I went with a different route since this test has multiple cases.

Emilgardis:
Maybe however, these tests should just check for existance or non-existance of `feature_*` in the output, to make it simpler and to minimize added lines

Enselic:
I took a quick look at the test code and it looks brilliant.

Just a quick note: Yes, maybe we should add a new small test API and test against that instead:, `test-apis/feature_gated_api`. I think that would make sense. It keeps the other APIs straightforward. And `example_api`, which would be another candidate, is mostly meant to be used for git tests, so I don't think we should feature gate that.

I wanted to mention this quickly before you got carried away, because I'm not sure I will have time for a detailed code review today. But from what I saw so far it looks great. Very nice trick with `BLESS` for example, on first sight.

test-apis/comprehensive_api/Cargo.toml
```diff
@@ -4,5 +4,9 @@ name = "comprehensive_api"
 version = "0.1.0"
 edition = "2021"
 
+[features]
```
Enselic: I think we can fully revert the changes to `comprehensive_api` now? "comprehensive" should be interpreted as "exercise all rustdoc JSON code processing paths in `public_api` crate", and for that we don't need items behind features. So if it's OK with you I'd like to keep `comprehensive_api` without feature flags.

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -560,6 +560,107 @@ fn initialize_test_repo(dest: &Path) {
...
+    )?;
+
+    let root = Path::new(
+        metadata
+            .pointer("/workspace_root")
+            .unwrap()
+            .as_str()
+            .unwrap(),
+    );
+
```
Enselic: If we change from `serde_json = "1"` to `cargo_metadata = "0.14.2"` (which is already in `Cargo.lock`) we can replace this code with

```suggestion
    let root = cargo_metadata::MetadataCommand::new()
        .exec()?
        .workspace_root;

```

which I think is a worthwhile thing to do. Note that specifying the version as just `"1"` is discouraged, because it is good practice to specify a lower bound on dependencies. See https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

Enselic (commented):
Looks very good. I just had two simple comments. After those are addressed I see no obstacles to approving and merging this.

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Emilgardis: done!

Emilgardis:
another tip (if you didn't already know, this feature is not very discoverable...)

press the link on `force-pushed`
image

to see a diff between commits that's been overwritten by force

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -560,6 +560,95 @@ fn initialize_test_repo(dest: &Path) {
...
+        for feature in features.features {
+            cmd.args(["--features", feature]);
+        }
+
+        if std::env::var("BLESS").is_ok() {
+            let out = cmd.output().unwrap();
+            std::fs::write(expected_file, out.stdout).unwrap();
+        } else {
+            // Make into a string to show diff
+            let expected = String::from_utf8(
```
Enselic, line 641: I think we can simplify with https://doc.rust-lang.org/std/fs/fn.read_to_string.html here?

But that is not going to stop me from merging this. Thank you so much for this PR.

Merging shortly.

Enselic (approved):

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Emilgardis: essentially the same thing, but clearer imo, I'll change to that if you want

Enselic:
I saw your comments on the RUSTUP_TOOLCHAIN  PR btw. I will experiment a bit more with that PR to make sure I understand things properly. Can't promise I will have time for that today though.

rustdoc-json/src/build.rs
```diff
@@ -39,34 +33,47 @@ pub(crate) fn run_cargo_rustdoc(options: BuildOptions) -> Result<PathBuf, BuildE
...
         command.arg(target);
     }
+    if *no_default_features {
+        command.arg("--no-default-features");
+    }
+    if *all_features {
+        command.arg("--all-features");
+    }
+    for feature in features {
+        command.args(["--features", &*feature]);
```
Enselic, line 75: @Emilgardis I'm a bit confused because on current `origin/main` with toolchain 1.63 I get this lint on this line:

```
error: deref on an immutable reference
  --> rustdoc-json/src/build.rs:80:37
   |
80 |         command.args(["--features", &*feature]);
   |                                     ^^^^^^^^^
   |
   = note: `-D clippy::borrow-deref-ref` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#borrow_deref_ref
help: if you would like to reborrow, try removing `&*`
   |
80 |         command.args(["--features", feature]);
   |                                     ~~~~~~~
help: if you would like to deref, try using `&**`
   |
80 |         command.args(["--features", &**feature]);
   |                                     ~~~~~~~~~~
```

but CI is also using 1.63, so it should also complain. The only difference is that I am on Apple Silicon while CI runs on Linux. But that shouldn't matter.

Do you also get this lint error with toolchain 1.63?

Emilgardis: hah, that's strange, yes I get the lint and it's correct. Not sure why it didn't triggered

Enselic: https://github.com/Enselic/cargo-public-api/pull/118 passed both locally for me and in CI, so I merged that

Emilgardis: Oh, it was triggered in CI

https://github.com/Enselic/cargo-public-api/runs/8049349007?check_suite_focus=true#step:3:137

Enselic: It was indeed! Thanks! Something is wrong with my CI script setup. I'll need to fix that :)

Enselic: (Fixed by https://github.com/Enselic/cargo-public-api/pull/120)

PR #117, opened 2022-08-26
Warn about clippy lints in code, leave deny to CI

Enselic:
We must not use `deny` in code because that will make e.g. `git bisect` or `--diff-git-checkouts` annoying in the future.

I probably missed to add `warn` in some place, but this is a good start.

CC @Emilgardis

### DIFF
--- a/cargo-public-api/src/lib.rs
+++ b/cargo-public-api/src/lib.rs
@@ -1,3 +1,6 @@
+// deny in CI, only warn here
+#![warn(clippy::all, clippy::pedantic)]
+
 /// For self-testing purposes. Please ignore.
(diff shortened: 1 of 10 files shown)


Emilgardis (approved):
Looks fine to me. Personally I dislike using pedantic and instead only add specific lints in pedantic directly

Enselic:
Thanks for review! I've enforced pedantic from day 1 of this project, and I find in general it results in better code. There are less than a handful `allow` in the code base where pedantic gave bad advice. I agree it occasionally is annoying. But I'd like to keep pedantic, because I think a code base that pass pedantic is nicer to work with than one that doesn't.

PR #119, opened 2022-08-27
Support `cargo +toolchain public-api` syntax

Enselic:
This is prototype quality code that is heavily influenced by https://github.com/Enselic/cargo-public-api/pull/113 and a large portion of credit should be given to @Emilgardis

@Emilgardis If we end up using this, I want to make sure you feel properly credited, since I was heavily influenced by your code. For example, I would be perfectly fine with assigning you the copyright of this code and committing it in your name. Let me know what you prefer.

Now too the issue at hand: After thinking some more about this problem, my conclusion is that we must keep `--tolchain` after all. Because `+toolchain` only works when we are invoked with the rustup proxy, i.e. as `cargo +toolchain public-api`. When we are invoked with `cargo run` in our git, and as a raw binary, i.e `cargo-public-api`, it must also be possible to specify a toolchain. Now that I think about it... maybe we should parse `+` ourselves? 🤔 My gut feeling is "no", but it can certainly be discussed.

In any case, would you mind taking a look at this draft code? It should pass all tests, and it `cargo +toolchain public-api` seems to work according to my quick testing. I will need to add proper automated tests if we decide to go this route.

Looking forward to your thoughts!

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -22,6 +22,15 @@ mod markdown;
 mod output_formatter;
 mod plain;
 
+#[derive(Debug)]
+struct App {
+    /// With what args we were invoked.
+    args: Args,
+
+    /// If we were invoked as a cargo subcommand.
+    as_subcommand: bool,
+}
+
 #[derive(Parser, Debug)]
 #[clap(author, version, about, long_about = None)]
 #[allow(clippy::struct_excessive_bools)]
@@ -114,9 +123,10 @@ pub struct Args {
 
     /// Allows you to build rustdoc JSON with a toolchain other than `+nightly`.
     /// Useful if you have built a toolchain from source for example, or if you
-    /// want to use a fixed toolchain in CI.
-    #[clap(long, default_value = "+nightly")]
-    toolchain: String,
+    /// want to use a fixed toolchain in CI. Note that you can also use the
+    /// regular toolchain override syntax `cargo +custom public-api`.
+    #[clap(long)]
+    toolchain: Option<String>,
 
     /// Build for the target triple
     #[clap(long)]
@@ -153,7 +163,8 @@ struct PostProcessing {
 }
 
 fn main_() -> Result<()> {
-    let args = get_args();
+    let app = get_app();
+    let args = app.args;
 
     let post_processing = if let Some(commits) = &args.diff_git_checkouts {
         print_diff_between_two_commits(&args, commits)?
@@ -277,16 +288,55 @@ impl Args {
     }
 }
 
+impl App {
+    fn rustdoc_json_toolchain(&self) -> String {
+        if self.as_subcommand {
+            // `RUSTUP_TOOLCHAIN` is expected to be set. Use it. But if we are
+            // run with the stable toolchain, use `+nightly`.
+            let rustup_toolchain =
+                std::env::var("RUSTUP_TOOLCHAIN").unwrap_or_else(|_| "nightly".to_owned());
+            let probably_stable = rustup_toolchain.starts_with("1.");
+            format!(
+                "+{}",
+                if probably_stable {
+                    "nightly"
+                } else {
+                    &rustup_toolchain
+                }
+            )
+        } else {
+            // invoked as a regular program. Use `--toolchain` since
+            // `RUSTUP_TOOLCHAIN` is not set, and default to `+nightly`.
+            self.args
+                .toolchain
+                .as_deref()
+                .unwrap_or("+nightly")
+                .to_owned()
+        }
+    }
+}
+
 /// Get CLI args via `clap` while also handling when we are invoked as a cargo
 /// subcommand
-fn get_args() -> Args {
+fn get_app() -> App {
     // If we are invoked by cargo as `cargo public-api`, the second arg will
     // be "public-api". Remove it before passing args on to clap. If we are
     // not invoked as a cargo subcommand, it will not be part of args at all, so
     // it is safe to filter it out also in that case.
-    let args = std::env::args_os().filter(|x| x != "public-api");
+    let mut args = vec![];
+    let mut as_subcommand = false;
+    for (index, arg) in std::env::args_os().enumerate() {
+        if arg == "public-api" && index == 1 {
+            as_subcommand = true;
+        } else {
+            args.push(arg);
+        }
+    }
 
-    Args::parse_from(args)
+    App {
+        args: Args::parse_from(args),
+        as_subcommand,
+    }
 }
 
 /// Figure out what [`Options`] to pass to
@@ -302,7 +352,8 @@ fn get_options(args: &Args) -> Options {
 /// `commit` is `Some` and thus a `git checkout` will be made, also return the
 /// original branch.
 fn collect_public_api_from_commit(commit: Option<&str>) -> Result<(PublicApi, Option<String>)> {
-    let args = get_args();
+    let app = get_app();
+    let args = &app.args;
 
     // Do a git checkout of a specific commit unless we are supposed to simply
     // use the current commit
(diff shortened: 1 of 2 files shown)


cargo-public-api/src/main.rs
```diff
@@ -277,16 +288,55 @@ impl Args {
...
+            )
+        } else {
+            // invoked as a regular program. Use `--toolchain` since
+            // `RUSTUP_TOOLCHAIN` is not set, and default to `+nightly`.
+            self.args
+                .toolchain
+                .as_deref()
+                .unwrap_or("+nightly")
+                .to_owned()
+        }
```
Emilgardis, line 315: my stable toolchain is called `stable-<target>` and according to this that would not be a stable compiler

example:
```
$ rustup toolchain install stable
$ rustup run stable env 
# prints RUSTUP_TOOLCHAIN=stable-*
```

this is what I mentioned [here](https://github.com/Enselic/cargo-public-api/pull/113#discussion_r956458156)

rustdoc-json/src/build.rs
```diff
@@ -57,6 +57,12 @@ fn cargo_rustdoc_command(options: &BuildOptions) -> Command {
     let overridden_toolchain = OVERRIDDEN_TOOLCHAIN.map(OsStr::new);
     if let Some(toolchain) = overridden_toolchain.or(requested_toolchain.as_deref()) {
         command.arg(toolchain);
+    } else {
+        // Remove `RUSTUP_TOOLCHAIN` so that it does not override that the caller
+        // of this function explicitly requested no specific toolchain, i.e.
+        // that the default toolchain shall be used. See
+        // <https://rust-lang.github.io/rustup/overrides.html>.
+        command.env_remove("RUSTUP_TOOLCHAIN");
```
Emilgardis, line 65: I'm confused.

This means that if `requested_toolchain` is `None`, and the binary where this is executed in is invoked with `RUSTUP_TOOLCHAIN=mytoolchain`, that fact will be lost and the default `rustup show active-toolchain` will be used instead.

Emilgardis:
I'd appreciate it if you, instead of doing a @mention, did a co-authored by

`Co-authored-by: Emil Gardström <emil.gardstrom@gmail.com>`

that also means less pings for me, as commits with @ghost mentions will end up in my inbox 😅, not a good experience if someone does a funky rebase or similar

rustdoc-json/src/build.rs
Enselic: Here is how I reason about it:

We need to keep in mind that this code runs in the context of a library. So if a client of the library asks to build rustdoc JSON with `toolchain: None` (read: default toolchain), then it is the responsibility of the library to ensure that happens. And to ensure that it happens, it needs to get rid of `RUSTUP_TOOLCHAIN`.

In the context of `cargo-public-api`, it is the responsibility of `cargo-public-api` to examine its environment, and depending on its environment, give proper instructions to the `rustdoc_json` library. Which is what happens in this PR.

cargo-public-api/src/main.rs
Emilgardis: From https://github.com/Enselic/cargo-public-api/commit/5cc0af2d3d0b8ee48fdde116f0113c0c9cdca3fc#r82373796, I deleted my comment on the commit and made it a review comment instead, sorry for the confusion.


> Aha that was what you meant before too. That can be solved with `rustup_toolchain.starts_with("1.") || rustup_toolchain.starts_with("stable")` I think?



you'd have to add `beta` to that too, and then also consider the case where a stable toolchain has been linked and would be named something completely different. There is a good reason to ask cargo what it is, instead of assuming by toolchain name

Enselic:
> I'd appreciate it if you, instead of doing a @mention, did a co-authored by
> 
> `Co-authored-by: Emil Gardström <emil.gardstrom@gmail.com>`
> 
> that also means less pings for me, as commits with @ghost mentions will end up in my inbox 😅, not a good experience if someone does a funky rebase or similar

Sorry! I should have known better 😅 Will fix in next push!

cargo-public-api/src/main.rs
Enselic: Ok, you have convinced me. In the next iteration of this PR (if we continue this route), I will use the approach in your PR. Thank you for the insights.

(I think we can can continue discussing the other aspects of this PR before I make an update to it though.)

rustdoc-json/src/build.rs
Emilgardis: so, consumers of `rustdoc_json` need to be aware that setting `toolchain = None` means that `RUSTUP_TOOLCHAIN` will not be respected by the library, and that further rustup shims called will also not respect it.

Enselic: Yes, that is correct. The way I see it, `RUSTUP_TOOLCHAIN` is not part of the API surface of `rustdoc_json`.

Nothing prevents us from making `RUSTUP_TOOLCHAIN` formally part of the API surface of course, but I think a library API that involves environment variables is an objectively badly designed API, because environment variables are leaky.

Emilgardis: What I'm trying to get at, and I think this is where our split is, is:

Suppose that I as a consumer of `rustdoc_json` is(are?) a cargo subcommand and I rely on the fact that people calling me as `cargo +nightly-2022-08-27 my-command` means `RUSTUP_TOOLCHAIN` will be set to `nightly-2022-08-27-<target>`. 
Commands like `rustc` will use that specific compiler and everything is well.
If I then use `rustdoc_json::build(BuildOptions::default())`, and then run `rustc` again, it'll maybe be a different rustc.

I don't see how `toolchain = None` would mean that `RUSTUP_TOOLCHAIN="" rustup show active-toolchain` is to be used, instead of the one already determined to be in use.

I'm instead suggesting that `toolchain = None` is equivalent to `toolchain = Some(std::env::var("RUSTUP_TOOLCHAIN"))`, but only by proxy of how rustup shims work

Emilgardis:
With this (and how the tests are currently), how do I test for example `print_public_api_with_blanket_implementations` with a specific nightly toolchain?

rustdoc-json/src/build.rs
Enselic: Good point, you're right! It is very evil of `rustdoc_json` to clear that environment variable, since it is important global state! I feel dumb for not realizing that earlier.

I *guess* we could restore the old value after we are done with invoking `cargo`, but that is racy, since we are dealing with global state.

If only there was a syntax for "use the default toolchain and ignore `RUSTUP_TOOLCHAIN`", like `cargo +default ...`. That would solve our problem. But I don't know of any such way.

If we re-design the API of `rustdoc_json` and make `RUSTUP_TOOLCHAIN` a formal part of the API, we still have the inconvenient situation of there being two toolchains involved at once. At least for the `cargo run` case. One toolchain to build the tool, and one to build the rustdoc JSON.

I'm not sure by now how to best proceed here...

Enselic:
> With this (and how the tests are currently), how do I test for example `print_public_api_with_blanket_implementations` with a specific nightly toolchain?

For that particular use case, `RUSTDOC_JSON_OVERRIDDEN_TOOLCHAIN_HACK=+custom cargo test print_public_api_with_blanket_implementations` should work

rustdoc-json/src/build.rs
Emilgardis:
```
Command::new("cargo").env_remove("RUSTUP_TOOLCHAIN")....
```

should work just fine

Enselic: Clever!

Another option would be to not make what toolchain to use optional, since `cargo +toolchain` overrides `RUSTUP_TOOLCHAIN`

But I like your solution more

Enselic: Hmm that code is already the code that was used... sorry... I am starting to think and read sloppily and rush too much... I should stop for today... thank you for good discussions so far.

Emilgardis: > Hmm that code is already the code that was used

what code? `.env_remove("RUSTUP_TOOLCHAIN")` only removes the var in the sub-process

Enselic:
https://github.com/Enselic/cargo-public-api/pull/113 has been merged 🎉

PR #126, opened 2022-09-01
Emit a warning when toolchain is overridden and not usable with `rustdoc_json`

### LINKED ISSUE #123: warn when toolchain is replaced with nightly
Some type of warning should probably be emited when the used toolchain, specified explicitly or not, is not used for rustdoc.

This due to lessen the belief that `cargo +stable public-api` actually works, even if it doesn't right now.

Emilgardis:
this pr resolves #123 

It does it in a tricky way, and I'm not sure I like.

Anyway, this also fixes a bug where you could do `cargo-public-api --toolchain +my-nightly` and we'd not use that toolchain, but `+nightly`.

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -152,6 +152,12 @@ fn main_() -> Result<()> {
 
     // check if using a stable compiler, and use nightly if it is.
     if active_toolchain_is_probably_stable(args.toolchain.as_deref()) {
+        if let Some(toolchain) = args
+            .toolchain
+            .or_else(active_toolchain_is_environment_override)
+        {
+            eprintln!("Warning: using the `{toolchain}` toolchain for gathering the public api is not possible, switching to `nightly`");
+        }
         args.toolchain = Some("+nightly".to_owned());
     }
 
@@ -189,6 +195,27 @@ fn active_toolchain_is_probably_stable(toolchain: Option<&str>) -> bool {
     };
 
     version.starts_with("cargo 1") && !version.contains("nightly")
+}
+
+/// returns the current toolchain if it was overriden by environment
+fn active_toolchain_is_environment_override() -> Option<String> {
+    let mut cmd = std::process::Command::new("rustup");
+    cmd.args(["show", "active-toolchain"]);
+    cmd.env_remove("RUSTUP_TOOLCHAIN");
+
+    let output = String::from_utf8(cmd.output().ok()?.stdout).ok()?;
+
+    output
+        .split(char::is_whitespace)
+        .next()
+        .and_then(|default| {
+            let toolchain = std::env::var("RUSTUP_TOOLCHAIN").ok();
+            if toolchain.as_deref() == Some(default) {
+                None
+            } else {
+                toolchain
+            }
+        })
 }
 
 fn check_diff(args: &Args, diff: &Option<PublicItemsDiff>) -> Result<()> {
--- a/scripts/test-invocation-variants.sh
+++ b/scripts/test-invocation-variants.sh
@@ -103,3 +103,16 @@ if ${cmd} >/dev/null 2>/dev/null; then
 else
     echo "PASS"
 fi
+
+if ! cargo +beta -v >/dev/null 2>/dev/null; then
+    rustup toolchain install beta --no-self-update
+fi
+
+cmd="cargo +beta public-api"
+echo -n "${cmd} ... "
+if n=$(${cmd} 2>&1 | grep "Warning: using the \`beta.*\` toolchain for gathering the public api is not possible"); then
+    echo "PASS"
+else
+    echo "FAIL: Using '+beta' to build rustdoc JSON should have mentioned the upgrade to `+nightly`"
+    exit 1
+fi


cargo-public-api/src/main.rs
```diff
@@ -118,7 +118,7 @@ pub struct Args {
     ///
     /// Useful if you have built a toolchain from source for example, or if you
     /// want to use a fixed toolchain in CI.
-    #[clap(long)]
+    #[clap(long, validator = |s: &str| if s.starts_with('+') { Ok(()) } else { Err("toolchain must start with a `+`")} )]
```
Enselic: nice 👍 

since this is not crucial (since there will be a follow-up error about missing toolchain), I think we can get away with not having an automated test for this, which normally is the case.

Enselic (commented):
Out of curiosity I took a peek, and admittedly I do not 100% understand what is going on (for that I would need to spend more time), but it does not look extremely horrible to me. Let me know when you have a solution that you personally also feel happy with, and I would be happy to take a closer look.

One high-level comment already: Wouldn't it be nice to move out the toolchain logic to a separate file `cargo-public-api/src/toolchain.rs`? So that `main.rs` becomes less "noisy", if you know what I mean. Maybe we could even have some `mod tests` in `toolchain.rs`. Not sure how much sense that would make, but maybe we could have some `Command::new("cargo")` tests there.

(did you get the mail I sent to your gmail address btw?)

Emilgardis:
I cant think of a better way to solve this. if this is not added, aa74a67 should be, and probably  54352ac too, since the changelog is pr based, I can make separate prs for them

cargo-public-api/src/main.rs
```diff
@@ -196,6 +205,35 @@ fn active_toolchain_is_probably_stable() -> bool {
...
+
+    let output = match cmd.output() {
+        Ok(output) => output,
+        Err(_) => return None,
+    };
+
+    let output = match String::from_utf8(output.stdout) {
+        Ok(version) => version,
+        Err(_) => return None,
+    };
```
Enselic: Can we de-duplicate this code somehow please? It is exactly the same above, and duplicated code quickly comes inconvenient to work with.

I think it could be nice to move the toolchain-related code (currently `active_toolchain_is_environment_override` and `active_toolchain_is_probably_stable`) into `src/toolchain.rs`.  Maybe even introduce a `struct Toolchain`? I can imagine that we'll find more cases in the future that requires even more logic, so I think it would be nice to isolate into a separate file.

Of these comments, doing the de-duplication is the most important, I think

cargo-public-api/src/main.rs
```diff
@@ -159,7 +159,13 @@ fn main_() -> Result<()> {
     let mut args = get_args();
 
     // check if using a stable compiler, and use nightly if it is.
-    if active_toolchain_is_probably_stable() {
+    if active_toolchain_is_probably_stable(args.toolchain.as_deref()) {
+        if let Some(toolchain) = args
+            .toolchain
+            .or_else(active_toolchain_is_environment_override)
+        {
+            eprintln!("Warning: using the `{toolchain}` toolchain for gathering the public api is not possible, switching to `nightly`");
```
Enselic, line 159: It would be great if we could somehow have a test for this scenario. We will definitely not regret writing one, because as we keep iterating on the toolchain code, all the tests that we will have written will be very valuable.  And the toolchain selection code is a good example of code that is tricky to get right, which also motivates writing tests.

Enselic (commented):
Thanks! I took a more detailed look now.

I'd say that much more important than having a clean implementation, is to make sure to have a good set of regression tests, because that will make it very easy to iteratively improve on the code.

cargo-public-api/src/main.rs
Emilgardis: I'm not sure how a test for this would work or be written without it having it's own job in the ci matrix or making the test install a "non-standard" stable toolchain

cargo-public-api/src/main.rs
Emilgardis: what would the Toolchain struct include? or is it just a newrapper/braid around a string?

for full toolchain specification i'd be something like https://github.com/cross-rs/cross/blob/7b04b6dbcbbe947a5e4eee83f9864172daa44667/src/rustc.rs#L85-L92 (- `platform`), all of which are overkill except channel

I considered moving it out but didn't see a benefit due to not being able to do unit tests, as that'd require installing and querying toolchains with rustup on `cargo test`, and that sounds like a bad idea

cargo-public-api/src/main.rs
Enselic: Yes the "magic" toolchain-picking parts are a bit tricky to test... And after I added CI tips (CI-EXAMPLES.md), regular users will also make use of it to lock against a fixed version of nightly. Since regular users use it, it makes it important functionality, which in turn justifies spending some extra time on getting a good testing infrastructure of it.

Since your work is entirely volunteered, I am tempted to say "let's merge this without tests", because I don't want to discourage you. But at the same time, not writing tests for this will make doing every upcoming release feel scary, because we won't know if this functionality broke. So I'm afraid I have to insist on somehow writing tests for this.

I currently have a lot on my plate, but I would be very happy with coming up with some good way to write tests. Right now it is not entirely clear for me on how to do it. But once I get time, I'm sure I can come up with something, if you find it hard to find the motivation for it yourself. I think it is very important that you consider it fun to hack on this project, and if you don't think it is fun to come up with test infrastructure, that's perfectly fine and reasonable :)

You mentioned creating a "ci matrix". For me "matrix" means [this](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs), but I'm not sure how we would make use of that for toolchain tests? Do you think you could elaborate?

Emilgardis: > Do you think you could elaborate?

I basically meant a ci job :).

something like

```yml
jobs:
  check-toolchain-warn:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: nightly
          profile: minimal
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: beta
          profile: minimal
      - run: cargo install --debug --path .
      - run: cargo +beta public-api -p cargo-public-api 2>&1 grep -q "switching to `nightly`"
```

could ofc append it to an existing job

Enselic: If that's what it takes, then that's what we will do. Just a few remarks:
* Couldn't we use an existing job and just install more toolchains in that job?
* We should ensure that ./scripts/run-ci-locally.sh will run the new stuff too

cargo-public-api/src/main.rs
Enselic: If you don't see any benefit of doing it, then we'll not do it. It was just some brainstorming :)

cargo-public-api/src/main.rs
Emilgardis: so I'll make `run-ci-locally.sh` install the other toolchain and then use that in say `test-invocation-variants.sh`?

Enselic: yep that sounds right

Emilgardis: i've implemented this, the reason for being a bit hard about it is that I don't really like the idea of installing random toolchains in tests, but I guess this will have to do!

Emilgardis:
should I split this pr up and only keep the last two commits here? since the first two commits are really fixes and should probably have a changelog entry

Enselic:
I'm a big fan of small and incremental work, so splitting it up into separate PRs that each individually does something sensible sounds great. It also makes review easier.

cargo-public-api/src/main.rs
Emilgardis: did some [changes](https://github.com/Enselic/cargo-public-api/compare/55bfd16382150d3c02907c932fbd3244d944a160..42ffde79c79abd822f555b09464e830af5d519c4) to not make it seem duplicated, and streamlined it a bit. I see a way to deduplicate it but it's pretty standard operation and we don't do too many cmd calls that needs it

Enselic (approved):
Looks good to me!

Emilgardis:
@Enselic I'm making a release, and in the notes it's mentioning to bump the [compatibility matrix](https://github.com/Enselic/cargo-public-api#compatibility-matrix), however I'm unsure what to change it to

```md
| cargo-public-api | Understands the rustdoc JSON output of  |
| ---------------- | --------------------------------------- |
| v0.15.0 - v0.16.x| nightly-2022-08-15 —                    |
| v0.13.x          | nightly-2022-08-10 — nightly-2022-08-14 |
| v0.12.x          | nightly-2022-05-19 — nightly-2022-08-09 |
| v0.10.x          | nightly-2022-03-14 — nightly-2022-05-18 |
| v0.5.x           | nightly-2022-02-23 — nightly-2022-03-13 |
| v0.2.x           | nightly-2022-01-19 — nightly-2022-02-22 |
| v0.0.5           | nightly-2021-10-11 — nightly-2022-01-18 |
```

maybe? also, what happened with 0.14.0?

edit: Or maybe I'm missing that the entries are for when the rustdoc json format version has changed, so no change needed for this release

Enselic:
@Emilgardis I should add this to MAINTAINER.md, but the bumping rules are:

x.0.0: We bump to 1.0.0 earliest when the rustdoc JSON format has stabilized, which probably will take many more months, maybe years

0.x.0: We bump it when `public_api` has had backwards incompatible changes, but I don't think we've done any such changes since v0.15.0? `git diff v0.15.0 -- public-api/public-api.txt` prints no diff. We also bump this when the rustdoc JSON format changes in a backwards incompatible way.

0.0.x: We bump it whenever we want to make a release but we don't have to/want to bump 0.x.0

Regarding Compatibility Matrix, yes it focuses on rustdoc JSON only, so to keep it simple I am omitting e.g. 0.14.0 from the list. I am only listing the "latest" release for a given nightly range.

Also, in `How to release` in MAINTAINER.md it says "push to `main`", but since this is your first release I think it is better that you create a PR with the release preparation. Then after I approve that, you are free to merge to main and then trigger a release from main.

Emilgardis:
there is a breaking change, the removal of output-format md no?

Enselic:
Ah yes, that justifies bumping to 0.16.0, I agree.

PR #129, opened 2022-09-03
correctly identify toolchain specified with `--toolchain`

Emilgardis:


Enselic (approved):
Would be nice with a regression test  for this (that fails without the fix), but I've nagged enough about tests before, so I'll let it slip this time. Maybe you'll add tests in another PR :)

(From now on whenever I approve a PR you are free to merge it when you see fit)

Emilgardis:
I agree with adding a test for this, this one feels better to add one for, and it could be a regular test, only need to add another invocation here https://github.com/Enselic/cargo-public-api/blob/a4945abbc58f0f6edbdabf0a928ab67cef0052f4/scripts/test-invocation-variants.sh#L89 that uses `--toolchain`

Emilgardis:
https://github.com/Enselic/cargo-public-api/runs/8171051195?check_suite_focus=true and the regression test works

Enselic (approved):
Great, thank you!

PR #132, opened 2022-09-04
test-invocation-variants.sh: Don't hardcode expected output

Enselic:
This is a preparation for an upcoming PR where the expected output will come from public-api lib.rs instead of cargo-public-api lib.rs.


Enselic:
Merging since I classify this as "obviously an improvement" and because it blocks other PRs

PR #134, opened 2022-09-04
test-invocation-variants.sh: Use central code to install all toolchains

Enselic:
And check what `MINIMUM_RUSTDOC_JSON_VERSION` is instead of hardcoding it.

CC @Emilgardis to avoid conflicts (maybe you are also changing this file)

FYI, I generally try to wait one day after creating a PR before I merge it, to give both me and potentially others some time to think of problems the PR might have. But if I am somewhat comfortable that the PR makes sense, I will merge it without review.

However, when I touch some major component that is contributed by someone else, I tend to request a code review. So for the rendering system I ping douweschulte, and for `--features`, +toolchain`, `--package` etc I will prefer to ping you for CR.

However, there is never any obligation or urgency to review, of course.

### DIFF
--- a/scripts/test-invocation-variants.sh
+++ b/scripts/test-invocation-variants.sh
@@ -1,6 +1,26 @@
 #!/usr/bin/env bash
 set -o errexit -o nounset -o pipefail
 
+# The oldest nightly toolchain that we support
+minimal_toolchain=$(sed -n 's/pub const MINIMUM_RUSTDOC_JSON_VERSION: &str = "\(nightly-[^"]\+\)";/\1/p' public-api/src/lib.rs)
+
+# A toolchain that produces rustdoc JSON that we do not understand how to parse.
+unusable_toolchain="nightly-2022-06-01"
+
+toolchains_to_install="
+    beta
+    $minimal_toolchain
+    $unusable_toolchain
+"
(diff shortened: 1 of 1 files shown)


scripts/test-invocation-variants.sh
```diff
@@ -1,6 +1,26 @@
 #!/usr/bin/env bash
 set -o errexit -o nounset -o pipefail
 
+# The oldest nightly toolchain that we support
+minimal_toolchain=$(sed -n 's/pub const MINIMUM_RUSTDOC_JSON_VERSION: &str = "\(nightly-[^"]\+\)";/\1/p' public-api/src/lib.rs)
```
Emilgardis, line 5: you could make the simple `public-api` output this with a flag, say `cargo run --bin public-api -- --minimal-rust-version`

Emilgardis (approved):
lgtm, minor nit about querying the `MINIMUM_RUSTDOC_JSON_VERSION`

scripts/test-invocation-variants.sh
Enselic: Good idea, the current solution is indeed a bit of a hack. If it breaks, I will do what you proposed.

PR #136, opened 2022-09-05
bump `rustdoc-types` to 0.15.0

Emilgardis:


Emilgardis:
this is a breaking change, as `public-api` exports types from that crate

Emilgardis:
aha, #135 uses this, closing this pr for that one

Enselic:
Yep, I have prepared for https://github.com/rust-lang/rust/pull/101386 and https://github.com/rust-lang/rust/pull/100762, but would be more than happy to let you take care of the rustdoc JSON format change(s) that comes after that.

PR #137, opened 2022-09-06
Add `--print-minimum-rustdoc-json-version` to `public-api`

Enselic:
And use it in `test-invocation-variants.sh`. The current hack is already broken for me 😳

Idea from https://github.com/Enselic/cargo-public-api/pull/134#discussion_r962331019


Emilgardis (approved):
haha

lgtm

PR #141, opened 2022-09-11
Properly render short-hand qualified paths with `Self::`

### LINKED ISSUE #140: Incorrect output for associated types in traits
With this code:

```rust
pub trait Trait {
    type Frobnicator;

    fn frobnicate(&self) -> Self::Frobnicator;
}
```

The output of `cargo public-api` is: (Note `<Self as >`)

```diff
pub fn public_api_test::Trait::frobnicate(&self) -> <Self as >::Frobnicator
pub mod public_api_test
pub trait public_api_test::Trait
pub type public_api_test::Trait::Frobnicator
```

I think the expected output should be: (Note `<Self as Trait>`)

```diff
pub fn public_api_test::Trait::frobnicate(&self) -> <Self as Trait>::Frobnicator
pub mod public_api_test
pub trait public_api_test::Trait
pub type public_api_test::Trait::Frobnicator
```

Enselic:
Closes #140

### DIFF
--- a/public-api/tests/expected-output/comprehensive_api.txt
+++ b/public-api/tests/expected-output/comprehensive_api.txt
@@ -80,6 +80,8 @@ pub fn comprehensive_api::structs::Plain::s3(&mut self)
 pub fn comprehensive_api::structs::Plain::s4(&'a self)
 pub fn comprehensive_api::structs::WithLifetimeAndGenericParam::new(unit_ref: &'b Unit, t: String) -> Self
 pub fn comprehensive_api::traits::Simple::act()
+pub fn comprehensive_api::traits::TraitReferencingOwnAssociatedType::own_associated_type_output(&self) -> Self::OwnAssociatedType
+pub fn comprehensive_api::traits::TraitReferencingOwnAssociatedType::own_associated_type_output_explicit_as(&self) -> <Self as TraitReferencingOwnAssociatedType>::OwnAssociatedType
 pub macro comprehensive_api::simple_macro!
 pub mod comprehensive_api
 pub mod comprehensive_api::attributes
@@ -147,7 +149,9 @@ pub trait comprehensive_api::traits::AssociatedConstDefault
 pub trait comprehensive_api::traits::AssociatedType
 pub trait comprehensive_api::traits::Empty
 pub trait comprehensive_api::traits::Simple
+pub trait comprehensive_api::traits::TraitReferencingOwnAssociatedType
 pub type comprehensive_api::traits::AssociatedType::Type
+pub type comprehensive_api::traits::TraitReferencingOwnAssociatedType::OwnAssociatedType
 pub type comprehensive_api::typedefs::RedefinedResult<T, E> = Result<T, E>
 pub type comprehensive_api::typedefs::TypedefPlain = Plain
 pub union comprehensive_api::unions::Basic
--- a/public-api/tests/expected-output/comprehensive_api.txt
+++ b/public-api/tests/expected-output/comprehensive_api.txt
@@ -82,6 +82,7 @@ pub fn comprehensive_api::structs::WithLifetimeAndGenericParam::new(unit_ref: &'
 pub fn comprehensive_api::traits::Simple::act()
 pub fn comprehensive_api::traits::TraitReferencingOwnAssociatedType::own_associated_type_output(&self) -> Self::OwnAssociatedType
 pub fn comprehensive_api::traits::TraitReferencingOwnAssociatedType::own_associated_type_output_explicit_as(&self) -> <Self as TraitReferencingOwnAssociatedType>::OwnAssociatedType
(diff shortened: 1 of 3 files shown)


public-api/tests/expected-output/comprehensive_api.txt
```diff
@@ -80,6 +80,9 @@ pub fn comprehensive_api::structs::Plain::s3(&mut self)
 pub fn comprehensive_api::structs::Plain::s4(&'a self)
 pub fn comprehensive_api::structs::WithLifetimeAndGenericParam::new(unit_ref: &'b Unit, t: String) -> Self
 pub fn comprehensive_api::traits::Simple::act()
+pub fn comprehensive_api::traits::TraitReferencingOwnAssociatedType::own_associated_type_output(&self) -> Self::OwnAssociatedType
```
Emilgardis, line 83: would it make sense to maybe render this as 

```suggestion
pub fn comprehensive_api::traits::TraitReferencingOwnAssociatedType::own_associated_type_output(&self) -> <Self as _>::OwnAssociatedType
```

harudagondi: In something like:

```rust
pub trait Trait {
    type Frobnicator;

    fn frobnicate(&self) -> Self::Frobnicator;
}
```

I would expect it would show `Trait::frobnicate(&self) -> Self::Frobnicator`, since it knows that `Self` is of `dyn Trait`, and `Trait` has an associated type of `Frobnicator`. No need to specify where `Frobnicator` came from.

However, for code like these:

```rust
pub trait FooTrait {
    type Frobnicator;

    fn frobnicate(&self) -> Self::Frobnicator;
}

pub trait BarTrait {
    type Borfer: FooTrait;

    fn borf(&self) -> Self::Borfer;
    fn frobnicate(&self) -> <Self::Borfer as FooTrait>::Frobnicator {
        self.borf().frobnicate()
    }
}
```

I would expect something like this to show up:

```
pub fn public_api_test::BarTrait::borf(&self) -> Self::Borfer
pub fn public_api_test::BarTrait::frobnicate(&self) -> <Self::Borfer as FooTrait>::Frobnicator
pub fn public_api_test::FooTrait::frobnicate(&self) -> Self::Frobnicator
```

But currently, it shows something like this:

```
pub fn public_api_test::BarTrait::borf(&self) -> <Self as >::Borfer
pub fn public_api_test::BarTrait::frobnicate(&self) -> <<Self as >::Borfer as FooTrait>::Frobnicator
pub fn public_api_test::FooTrait::frobnicate(&self) -> <Self as >::Frobnicator
```

I think `Self::Borfer` and `Self::Frobnicator` is clearer and more concise than `<Self as _>::*`.

harudagondi (approved):

public-api/tests/expected-output/comprehensive_api.txt
Emilgardis: alright, I agree with you.

test-apis/comprehensive_api/src/traits.rs
```diff
@@ -16,6 +16,21 @@ pub trait AssociatedType {
...
+        &self,
+    ) -> <Self as TraitReferencingOwnAssociatedType>::OwnAssociatedType;
+}
+
+pub trait TraitWithGenerics<T, U> {
+    type Foo;
+
+    fn bar() -> <Self as TraitWithGenerics<T, U>>::Foo;
+}
+
```
Emilgardis, line 33: There's currently no usage of a trait with an associated type in the api tests. Could we add that?

```rs

pub struct FooBar;

impl TraitReferencingOwnAssociatedType for FooBar {
	type OwnAssociatedType = ();
    ...
}

impl TraitWithGenerics<(), ()> for FooBar {
    type Foo = ();
	...
}
```

for example

Emilgardis (changes requested):

public-api/tests/expected-output/comprehensive_api.txt
Enselic: Thanks for review! IMHO, in order for `cargo public-api` output to look familiar to as many users as possible, our output should be the same as rustdoc HTML. And rustdoc HTML renders it as `Self::OwnAssociatedType`, so I think we should do that too.

test-apis/comprehensive_api/src/traits.rs
Enselic: @Emilgardis Thanks for review! Would you be OK with you doing a follow-up PR that adds these tests? I think that is the best way to get the tests exactly as you want them :) I would be happy to review such a PR.

If that sounds OK, feel free to merge this PR on my behalf.

If you prefer me to add such tests before this PR is merged, that's also perfectly fine. Just let me know what you prefer.

Emilgardis: sounds good

Emilgardis (approved):

Issue #143, opened 2022-09-13
"error: no such subcommand: `+nightly-2022-09-07`" when using rustdoc-json on windows


jnises:
It seems I'm running into this bug https://github.com/rust-lang/rustup/issues/3036
Nested cargo invocations have issues in recent versions of rustup on windows.

I'm trying to run code like:
```rust
rustdoc_json::build(BuildOptions::default().toolchain("+nightly"))
```

I'm on Windows 11

I can work around the issue by making `cargo_rustdoc_command` call `rustup run nightly-2022-09-07 cargo ...` instead of `cargo +nightly-2022-09-07 ...`

What is the correct way to run nested a cargo with a different toolchain?

Enselic:
Thanks for reporting!

@Emilgardis Do you see any problem with us using `rustup run ...` instead?

@jnises Any chance you can create a PR with what works for you? If all our existing regression tests pass after the change, I am comfortable to use the change, because we have a good set of regression tests. 

Assuming @Emilgardis does not know of any issues with switching to that. But I do believe that is also what cargo-expand is using, probably for good reason.

Emilgardis:
switching to `rustup run <toolchain> cargo ...` when an override is mentioned sounds good, that's what +toolchain is supposed to be semantically equivalent to

Basically https://github.com/Enselic/cargo-public-api/blob/e40061548e90a96c639a7506939b74965677abc2/rustdoc-json/src/build.rs#L50-L56

would become

```rust
    let mut command = if let Some(toolchain) = OVERRIDDEN_TOOLCHAIN.or(requested_toolchain.as_deref()) {
        Command::new("rustup").args(["run", toolchain, "cargo"])
    } else {
        Command::new("cargo")
    }

    command.arg("rustdoc");    
```

jnises:
https://github.com/Enselic/cargo-public-api/pull/147 seems to do the same thing as my workaround.
So that should fix my issue :)

Enselic:
Closed by https://github.com/Enselic/cargo-public-api/pull/147

PR #147, opened 2022-09-14
fix windows `+toolchain` invocation

Emilgardis:
also adds caching to the test jobs and runs tests on all available platforms

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,15 +40,23 @@ jobs:
       - run: scripts/cargo-clippy.sh
 
   cargo-test:
-    name: cargo test
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
+    name: cargo test (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly
           profile: minimal
+      - uses: Swatinem/rust-cache@v2
       - run: cargo test --locked
+        env:
+          RUSTFLAGS: --cfg in_ci
 
   check-public-apis:
     name: Check public APIs
@@ -63,12 +71,20 @@ jobs:
           scripts/check-public-apis.sh
 
   test-invocation-variants:
-    name: scripts/test-invocation-variants.sh
-    runs-on: ubuntu-latest
+    name: scripts/test-invocation-variants.sh (${{ matrix.os }})
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly
           profile: minimal
+      - uses: Swatinem/rust-cache@v2
       - run: scripts/test-invocation-variants.sh
+        shell: bash
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -25,7 +25,9 @@ use test_utils::rustdoc_json_path_for_crate;
 #[path = "../src/git_utils.rs"] // Say NO to copy-paste!
 mod git_utils;
 
+// FIXME: This tests is ignored in CI due to some unknown issue with windows
 #[test]
+#[cfg_attr(all(target_family = "windows", in_ci), ignore)]
 fn list_public_items() {
     let cmd = Command::cargo_bin("cargo-public-api").unwrap();
     assert_presence_of_own_library_items(cmd);
@@ -51,7 +53,9 @@ fn custom_toolchain() {
     assert_presence_of_own_library_items(cmd);
 }
 
+// FIXME: This tests is ignored in CI due to some unknown issue with windows
 #[test]
+#[cfg_attr(all(target_family = "windows", in_ci), ignore)]
 fn list_public_items_explicit_manifest_path() {
     let mut cmd = Command::cargo_bin("cargo-public-api").unwrap();
     cmd.arg("--manifest-path");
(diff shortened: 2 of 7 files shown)


Emilgardis:
will implement the changes suggested in #143 after confirming if this fails or not.

If it does not fail, I'm not sure of a good way to replicate it, possibly adding a temporary file override for the directory is the best way.

Edit:

It works, running `cargo test` on 815cb57 fails on `custom_toolchain`

.github/workflows/CI.yml
```diff
@@ -40,14 +40,21 @@ jobs:
...
+          - windows-latest
+    name: cargo test (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly
           profile: minimal
+      - uses: Swatinem/rust-cache@v2
```
Emilgardis, line 56: since this is on nightly, is it even worth having a cache?

Emilgardis:
hmm, on my real windows machine this current test suite works.

```
---- list_public_items stdout ----
thread 'list_public_items' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 193, kind: Uncategorized, message: "%1 is not a valid Win32 application." }', cargo-public-api\tests\cargo-public-api-bin-tests.rs:521:25

---- list_public_items_explicit_manifest_path stdout ----
thread 'list_public_items_explicit_manifest_path' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 193, kind: Uncategorized, message: "%1 is not a valid Win32 application." }', cargo-public-api\tests\cargo-public-api-bin-tests.rs:521:25
```

Emilgardis:
https://github.com/Enselic/cargo-public-api/actions/runs/3056083057/jobs/4929869777
https://github.com/Enselic/cargo-public-api/actions/runs/3056238540/jobs/4930191239
https://github.com/Enselic/cargo-public-api/actions/runs/3056238540/jobs/4930191239

I don't understand why it's only `list_public_items` and `list_public_items_explicit_manifest_path` that fail. If these are failing, then `custom_toolchain` should fail to, no? and why does it work on my machine?

Very strange...

.github/workflows/CI.yml
Enselic: Only rustdoc JSON is built with nightly. Tests, libs and bins are built with stable, so I think it makes sense to use a cache. Thanks!

We can at least try, and if there is no difference from before, we can remove it again.

.github/workflows/CI.yml
```diff
@@ -63,12 +69,20 @@ jobs:
           scripts/check-public-apis.sh
 
   test-invocation-variants:
-    name: scripts/test-invocation-variants.sh
+    name: scripts/test-invocation-variants.sh (${{ matrix.os }})
```
Enselic, line 74: Running tests on all platforms is great 👍

Shoutout to GitHub for making this so easy to set up. Free Windows and Mac runners ❤️ :)

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -203,6 +211,7 @@ fn deny_changed_when_not_diffing() {
 }
 
 #[test]
+#[cfg_attr(target_family = "windows", ignore)]
```
Enselic, line 218: How come we ignore these tests but not the other two failing ones?

rustdoc-json/src/build.rs
```diff
@@ -47,11 +47,15 @@ fn cargo_rustdoc_command(options: &BuildOptions) -> Command {
...
     } = options;
-    let mut command = Command::new("cargo");
 
-    if let Some(toolchain) = OVERRIDDEN_TOOLCHAIN.or(requested_toolchain.as_deref()) {
-        command.arg(toolchain);
-    }
+    let mut command =
+        if let Some(toolchain) = OVERRIDDEN_TOOLCHAIN.or(requested_toolchain.as_deref()) {
+            let mut cmd = Command::new("rustup");
+            cmd.args(["run", toolchain.trim_start_matches('+'), "cargo"]);
```
Enselic, line 54: Maybe it would make more sense to change to an interface where toolchains are specified without +, so that we don't have to trim + here?

Enselic (commented):
Thank you for working on this. The failures are indeed very strange, especially if they work locally for you. I have no idea on how to fix it. I have no access to a Windows machine myself, unfortunately.

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Emilgardis: These tests are running a bash script. I tried to be smart by running these scripts as `bash ...` but that failed miserably on gh actions and locally.

this is the script https://github.com/Enselic/cargo-public-api/blob/1f77f1161def74b73c65021f5292535dbcfcc896/cargo-public-api/tests/cargo-public-api-bin-tests.rs#L497

rustdoc-json/src/build.rs
Emilgardis: 100%, kept this in to preserve compability

.github/workflows/CI.yml
Emilgardis: ofcourse! forgot `actions-rs/toolchain` doesn't override by default

Emilgardis:
this should be ready now

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Enselic: Looks like bash should be available in GitHub windows-latest runner: https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#shells

But we don't need to solve that in this PR, we can fix that later, especially if you already tried and it was tricky 👍

rustdoc-json/src/build.rs
Enselic: Ok, let's do that in a follow-up PR by someone at some point then 👍

Enselic (approved):
LGTM, thanks! There are some follow-up, optional polishing we can do, but nothing that blocks this PR from being merged, by my estimation

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Emilgardis: definitely, the problem is that gitbash get's overrided in path by wslbash, and if we use wsl bash, you'd need to have a wsl image installed, which the runner doesn't have by default.

Discovering the location of `gitbash` sounds kinda iffy also

Enselic:
@Emilgardis Any particular reason you are waiting with merging this? If you want to do something more first, no rush of course, just mainly wanted to make sure it is not just about that you simply forgot it :)

Emilgardis:
Oh sorry, no I thought you were going to hit merge :)

Enselic:
Closes #143

Issue #151, opened 2022-09-18
trait supertrait is not rendered


Emilgardis:
the following code

```rust
pub trait T: private::Sealed + Send {}

impl T for () {}

mod private {
    pub trait Sealed {}
    impl Sealed for () {}
}
```

generates

```
pub mod krate
pub trait krate::T
```

the supertrait `private::Sealed + Send` should be rendered here.

one side note also, the implementor `T for ()` should be rendered, is there an issue for that available?

Emilgardis:
Ideally, we'd also render `impl Sealed for () {}` as well

Enselic:
👍

I have started working on also rendering impls, but it is not ready yet: https://github.com/Enselic/cargo-public-api/pull/152

Emilgardis:
> Ideally, we'd also render `impl Sealed for () {}` as well

This is currently not implemented, should a new issue be created for that maybe?

Enselic:
I'm thinking that's part of #36, even though the title does not perfectly match.

I am currently prototyping what the code should look like to render all `impl`s in a public API. Right now it looks like each `PublicItem` will (pseudocode) get a field `impls: Vec<PublicItem>` and `children: Vec<PublicItem>` so that for a `pub struct Foo` the `cargo public-api` output would be

```rust
pub struct Foo
impl Display for Foo {      // <-- in `impls` for `PublicItem` for `Foo`
    fn fmt(etc)             // <-- in `children` for `PublicItem` for the `impl`
}
```

and in the future we could use `children` to render struct fields as part of structs:

```rust
pub struct Foo {
    pub a_field: bool,
}
```

The above feels promising to me even though this is only at early stages of experimentation.

Your case with `impl Sealed for () {}` is extra tricky since I don't think the rustdoc JSON contains an item for `()` with its list of `impl`s. But I think `Sealed` contains a list of `implementors`. But maybe we need to create a follow-up ticket later.

The above work is intended to also make progress on or fix #37 and #39

Enselic:
After https://github.com/Enselic/cargo-public-api/pull/195 lands we will start to render impls for (). Using the code in that PR, here is the public API listing for

```rust
pub trait T: private::Sealed + Send {}

impl T for () {}

mod private {
    pub trait Sealed {}
    impl Sealed for () {}
}
```

```
$ cargo run -- --manifest-path ~/src/lib/Cargo.toml --unsorted 
pub mod api_test
[...]

Issue #39, opened 2022-02-13
Present associated items as part of `impl`s


Enselic:
If a `struct Foo` implements both e.g. `TryFrom` and `TryInto`, the output will contain

```
pub type some_crate::Foo::Error
pub type some_crate::Foo::Error
``` 

This is because both `TryFrom` and `TryInto` has an associated type called `Error`.

The output should be more specific about where the `Error` comes from in each case, to make it clear it is not a matter of duplication.

Enselic:
cargo doc HTML does not present these in isolation, but as part of `impl`s. We show this:

```bash
% cd test-apis/example_api-v0.1.0
% cargo public-api --with-blanket-implementations
[...]
pub type example_api::Struct::Error = <U as TryFrom<T>>::Error
pub type example_api::Struct::Error = Infallible
```

i.e. the associated types are by themselves. That doesn't make them very "associated".

If we compare to cargo doc HTML, it looks like this:

```
% cargo doc --open
```
<img width="820" alt="Skärmavbild 2022-07-03 kl  21 15 38" src="https://user-images.githubusercontent.com/115040/177054302-0adaccbf-2e5b-4894-ba9d-53af495d97ea.png">

If you minimize an `impl`, the `Error` also goes away, and the `impl` has a larger font. Clearly (but perhaps not obviously) the `Error` is presented as part of an `impl`. So I think the right way to solve this is to also include `impl`s in our output. `impl`s are also items, after all.

This is highly related to 
* #36 
* #38

Issue #36, opened 2022-04-25
Show Impls instead of all functions


douweschulte:
As part of the tree data structure I was thinking about showing the `impl`s of a `struct`/`enum`/`trait` instead of the full list of functions. For local traits the functions will already be in the list (in the trait definition) and for other traits the functions can be assumed to be known or at least less part of this crates api and more for that other crate. Also this should remove clutter if only a `trait` is updated this should only list this change one time and not for every entity implementing it.

Implementing this change would be a nice first step for a more tree like data structure as discussed before. I have already tried some things locally but nothing is working yet.

Mock up:
```rust
#[derive(Clone, Copy, Debug)]
pub struct Plain {
    pub x: usize,
}
```
Output:
```
pub struct comprehensive_api::structs::Plain
Implements: Clone, Copy, Debug
```

Enselic:
I completely agree that it very well could make sense to show what traits are implemented rather than what functions (of traits) that are implemented. I'm not sure either exactly how to present it or diff it. But please continue experimenting and let me know if you figure out a good way :)

repi:
this would indeed be great, and important also for things like empty marker traits that do not have any functions or types in them, but if a type derives them or not is still semver-relevant and affect the public API.

Enselic:
I just merged https://github.com/Enselic/cargo-public-api/pull/152, so from now on, auto trait implementations are part of the output. These are a simple special case of `impl` rendering since there is never any associated items, so auto traits can always be rendered on one line as one item.

Enselic:
https://github.com/Enselic/cargo-public-api/pull/195 makes the output contain impls themselves (and their associated items), rather than "cherry-picking" only the associated items of impls.

The proposed expected output `Implements: Clone, Copy, Debug` will not work in the general case unfortunately, since impls can be parameterized by both lifetimes and generic args and generic bounds. So for full public API fidelity, we need to include the `impl` items themselves in the output.

The way I see it, https://github.com/Enselic/cargo-public-api/pull/195 will close this issue. But I am more than happy to discuss and iterate on the details.

douweschulte:
I agree this seems like a nice implementation. The proposed syntax indeed does not solve the general case.

PR #152, opened 2022-09-18
Include Auto Trait Implementations in API listings (and diffs)

Enselic:
This is important, because it is easy to accidentally break semver compatibility with subtle auto trait impl changes. Read https://predr.ag/blog/toward-fearless-cargo-update/ for example.

I considered not showing these by default, and instead adding a `--with-auto-trait-implementations` as an opt-in, but I think that becomes too complicated. Especially for CI uses cases where you would have to pass to many and too long flags to fully list and diff an API.

So instead, this PR always shows them. If it end up becoming too noisy, my plan is to get rid of `--with-blanket-implementations` and instead making the default output complete, and then add an opt-in to not include blanket and auto trait impls, called something like `--terse` or `--simplified`.

This is taking one step towards solving #36 and #39 and I think #151.


Enselic:
I _think_ this is ready for review, if anyone is interested. I would like to compare with rustdoc HTML output some more, and also probably write some more tests, but I have already done some iterations of this code, and consider it to be in quite good shape already. Might take a few more days before I get time to do the finishing touches.

Rendering Auto Traits impls is an easy special case, because they never have associated items, so they can always be rendered on one row. To render all kinds of impls, we need to think through how to render associated items. With indentation, maybe? Some experimentation is needed. To be done in follow-up PRs.

public-api/tests/expected-output/comprehensive_api.txt
```diff
@@ -2,6 +2,126 @@
...
+impl UnwindSafe for comprehensive_api::exports::issue_145::external_3::External
+impl UnwindSafe for comprehensive_api::structs::Plain
+impl UnwindSafe for comprehensive_api::structs::PrivateField
+impl UnwindSafe for comprehensive_api::structs::TupleStructDouble
+impl UnwindSafe for comprehensive_api::structs::TupleStructDoubleWithHidden
+impl UnwindSafe for comprehensive_api::structs::TupleStructDoubleWithPrivate
+impl UnwindSafe for comprehensive_api::structs::TupleStructSingle
+impl UnwindSafe for comprehensive_api::structs::Unit
+impl UnwindSafe for comprehensive_api::unions::Basic
+impl<'a, T, D> RefUnwindSafe for comprehensive_api::enums::EnumWithGenerics where D: RefUnwindSafe, T: RefUnwindSafe
```
Enselic: Oops, I was too sloppy when reviewing the output diff, because a regression slipped in here. rustdoc HTML output has generics params on the struct, i.e. it should be `EnumWithGenerics<'a, T, D>`. Full HTML output is:

```
impl<'a, T, D> RefUnwindSafe for EnumWithGenerics<'a, T, D> where D: RefUnwindSafe, T: RefUnwindSafe,
```

I have pushed a fix for this in 3aa688e.

Edit: Oops again, I meant to write this comment on https://github.com/Enselic/cargo-public-api/pull/156 ...

Enselic:
I have quite carefully compared new output with HTML output, and they are identical as far as I can tell. Merging.

PR #153, opened 2022-09-18
Replace `create-test-git-repo.sh` with Rust so tests run on Windows

Enselic:
With some luck this will also pass in Windows CI 🤞 (I wrote this code on a Mac)


Emilgardis:
quick shoutout to https://docs.rs/xshell/latest/xshell/ and the [`xtask`](https://github.com/matklad/cargo-xtask/) solution

Enselic:
Thanks for the tips! I've heard loosely about them before, but I try to stay as vanilla as possible for as long as possible, since I like things to be as simple as possible. Not least because it lower the barrier for new contributors.

But when things start to get out of hands, those are definitely things to consider.

In this case I think the Rust version doesn't look too bad (and now CI passes also on Windows for many more tests 🎉 )

Emilgardis:
if we gate `diff_public_items_with_dirty_tree_fails` so it's only ran on linux, we can enable macos too now for the tests

I think that was the only reason I didn't include macos in cargo test

Enselic:
Hmm the test passes on my Mac though. Let's first see what happens if we simply just add mac.

Emilgardis:
> Hmm the test passes on my Mac though. Let's first see what happens if we simply just add mac.

Yes mine too :3

Enselic:
The mac failure seems to have been caused by a different git version or config that printed a shorter error message to stdout. I changed the assert (to something that I think is also ok) and now the test pass also in Mac CI and still on my local Mac.

Issue #145, opened 2022-09-13
Privately renamed items can be visible in the public API listing


repi:
Ran into multiple problems regarding `use` statements on types (not not `pub use`) where the public API listing will be referring internal naming of the type within the module rather than the public type it is actually referring to. 

This input:

```rust
use macaw as internal_name_for_a_crate;
use macaw::Conformal3;
use macaw::Conformal3 as WeirdName;

pub fn test1(_transform: Conformal3) {}
pub fn test2(_transform: WeirdName) {}
pub fn test3(_transform: macaw::Conformal3) {}
pub fn test4(_transform: internal_name_for_a_crate::Conformal3) {}

pub enum Test {
    V1(Conformal3),
    V2(WeirdName),
    V3(macaw::Conformal3),
    V4(internal_name_for_a_crate::Conformal3),
}
```

Generates this output:

```rust
pub enum a2::Test
pub enum variant a2::Test::V1(Conformal3)
pub enum variant a2::Test::V2(WeirdName)
pub enum variant a2::Test::V3(macaw::Conformal3)
pub enum variant a2::Test::V4(internal_name_for_a_crate::Conformal3)
pub fn a2::test1(_transform: Conformal3)
pub fn a2::test2(_transform: WeirdName)
pub fn a2::test3(_transform: macaw::Conformal3)
pub fn a2::test4(_transform: internal_name_for_a_crate::Conformal3)
pub mod a2
```

Which I believe needs to be this to reflect what the public API actually is and not "leak" any internal naming within the crate module.

```rust
pub enum a2::Test
pub enum variant a2::Test::V1(macaw::Conformal3)
pub enum variant a2::Test::V2(macaw::Conformal3)
pub enum variant a2::Test::V3(macaw::Conformal3)
pub enum variant a2::Test::V4(macaw::Conformal3)
pub fn a2::test1(_transform: macaw::Conformal3)
pub fn a2::test2(_transform: macaw::Conformal3)
pub fn a2::test3(_transform: macaw::Conformal3)
pub fn a2::test4(_transform: macaw::Conformal3)
pub mod a2
```

Rustdoc HTML handles the the function output here correctly referring to all of them as `Conformal3` but all linking to the `macaw::Conformal3` docs page. I think in our case we would want it to be directly fully qualified names of  `macaw::Conformal3` everywhere where used.

[...]

PR #154, opened 2022-09-20
Add `RenderingContext` to `render.rs`

Enselic:
This used to be just a regular arg that was passed around but was later removed (see #78). It turns out we need something like this after all. Add it back, but instead of explicitly passing it around, pass it round implicitly via `self` in a new `RenderingContext`. This makes it a bit easier to work with.

This allows us to get rid of `pre_resolved_fields` since we can look it up on-demand in the render code. It also enables us (with some more upcoming work that I have prototyped) to solve both #145 and #36. Solving these issues requires us to resolve the path of items from only their ID.

### DIFF
--- a/public-api/src/intermediate_public_item.rs
+++ b/public-api/src/intermediate_public_item.rs
@@ -1,9 +1,8 @@
-use crate::render;
 use std::rc::Rc;
 
 use rustdoc_types::{Id, Item, Type};
 
-use crate::tokens::Token;
+use crate::{render::RenderingContext, tokens::Token};
 
 /// This struct represents one public item of a crate, but in intermediate form.
 /// It wraps a single [Item] but adds additional calculated values to make it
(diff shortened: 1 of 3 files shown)


Enselic:
@douweschulte Are you interested in reviewing this since it touches the heart of the rendering code?

Conceptually it is a simple change: Give the render code access to the `Crate` it renders, like you predicted would be needed in your initial code. (You get a nicer diff to review if you ignore whitespace.)

douweschulte (approved):
No problem. I do like the fact that the `RenderingContext` makes it conceptually clear that everything depends on the `Crate` reference. As you said it is conceptually a simple change, besides adding in `self` everywhere nothing much has changed. I approve.

As a side note I wanted to see if clippy had things to add about this change, it did not, but there where other things that clippy listed. I used the 'ultimate' clippy command: `cargo clippy -- -W clippy::nursery -W clippy::pedantic` I will work through them in a different PR to see if you agree with clippy.

PR #155, opened 2022-09-20
Pedantic clippy lints

douweschulte:
I followed pretty much all clippy lints (one was a false lint). I have the feeling these all help in cleaning up the code a tiny bit more. If you do not agree, or want some of them reverted just let me know.
```
cargo clippy -- -W clippy::nursery -W clippy::pedantic
```


douweschulte:
The failing tests seem to be because some functions changed from `fn` to `const fn` which changes the API. If you do not want to change the API in this way I can revert them. While technically a change in the API this should not break any code, only when reverting back to `fn` can this actually break code.

cargo-public-api/src/main.rs
```diff
@@ -178,13 +178,14 @@ fn main_() -> Result<()> {
...
 /// more info of how different toolchains can be activated.
 fn active_toolchain_is_probably_stable(toolchain: Option<&str>) -> bool {
-    let mut cmd = if let Some(toolchain) = toolchain {
-        let mut cmd = std::process::Command::new("rustup");
-        cmd.args(["run", toolchain.trim_start_matches('+'), "cargo"]);
-        cmd
-    } else {
-        std::process::Command::new("cargo")
-    };
+    let mut cmd = toolchain.map_or_else(
```
Enselic, line 183: I agree this PR as a whole is an overall improvement, but I'm unsure if the `map_or_else` changes actually improves readability. @Emilgardis Which version of this (and more below) code of yours do you prefer?

rustdoc-json/src/build.rs
```diff
@@ -186,14 +188,14 @@ impl BuildOptions {
 
     /// Whether to pass `--no-default-features` to `cargo rustdoc`. Default: `false`
     #[must_use]
-    pub fn no_default_features(mut self, no_default_features: bool) -> Self {
+    pub const fn no_default_features(mut self, no_default_features: bool) -> Self {
```
Enselic, line 195: My gut feeling is to keep the API as is, but maybe that's because I never made heavy use of the const stuff. @Emilgardis Do you have any preference here? The base code is also from you :)

cargo-public-api/src/main.rs
Emilgardis: it looks a bit strange, but I guess it's fine

rustdoc-json/src/build.rs
Emilgardis: going `fn` -> `const fn` is never a breaking change unless const generics are used (which they aren't here), doing this as much as possible is good because of being able to tell a consumer that "this function doesn't do anything strange, no heap allocations nor io" and it could enable further optimizations

I like it :)

douweschulte: One point against is that it will be harder to change it back (that will be a breaking change) so whenever these functions will be changed in the future everything inside will need be be kept const. But as there is not much floating point math going on in the project that should be possible (that was a major point in another crate I worked with). More and more stuff in the language itself is moving to const. Which can be seen as a argument to start using it here as well.

cargo-public-api/src/main.rs
Enselic: Ok, let's keep this change then.

rustdoc-json/src/build.rs
Enselic: Let's be adventurous and try this change out :)

Before 1.0.0 it is not a big deal to revert this change later if needed, and 1.0.0 is probably quite far into the future.

Enselic (approved):
Let's go ahead with this then!

To make CI pass I think you only need to run

    ./scripts/bless-public-apis.sh && ./scripts/bless-expected-output-for-tests.sh

and then squash and force push to this PR. You can confirm locally that CI will pass with `./scripts/run-ci-locally.sh`, if you want.

Feel free to merge this once CI passes.

rustdoc-json/src/build.rs
Emilgardis: I agree, non of these constifying measures are problematic imo for backward compatibility, as they all do stuff that clearly should not do io or other non-const operations

Enselic:
@douweschulte I hope I didn't discourage you by labeling this as "exclude-from-release-notes". If you want to be in the releases notes with this PR that is prefectly fine. We can just remove the label again :)

In general I try to restrict the (auto-generated) release notes to entries that affects end-users, but I should probably be less "aggressive" in excluding contributions from others!

douweschulte:
I have no problem with being excluded from release notes. But I did have problems finding a free evening ;-). And I do have problems running your scripts.
```
(base) douwe@desktop-douwe:/mnt/c/Users/douwe/src/cargo-public-api$ ./scripts/bless-public-apis.sh && ./scripts/bless-expected-output-for-tests.sh
: invalid option namec-apis.sh: line 2: set: errexit
./scripts/bless-public-apis.sh: line 3: $'\r': command not found
./scripts/bless-public-apis.sh: line 4: syntax error near unexpected token `$'do\r''
'/scripts/bless-public-apis.sh: line 4: `for crate in public-api rustdoc-json; do
(base) douwe@desktop-douwe:/mnt/c/Users/douwe/src/cargo-public-api$ ./scripts/run-ci-locally.sh
: invalid option namelly.sh: line 2: set: xtrace
```
Running on Ubuntu 20.04 WSL. If you have some ideas about how to fix it I would be happy to hear, otherwise I will spend some time debugging.

Emilgardis:
looks like you have the repo checked out with crlf, i.e windows line endings?

try (after commiting/stashing your changes so no work is lost)

```
git config core.autocrlf input
git add --renormalize .
```

douweschulte:
That was indeed part of the problem, the given solution did not help but `dos2unix` did. But now there is a new very interesting error: 
```
(base) douwe@desktop-douwe:/mnt/c/Users/douwe/src/cargo-public-api$ ./scripts/bless-expected-output-for-tests.sh
error: `rustc -vV` didn't have a line for `host:`, got:
```
I even updated cargo to see if that helped but that did do anything. I got the same error message while trying to install the package using cargo `cargo install cargo-public-api`. (All on ubuntu WSL).

But the bless api was easy enough to replicate on my windows machine which did not have these problems. The tests output however looked more daunting. If one of you can run that script that would help me. Otherwise I will translate the bash to a windows file later.

Emilgardis:
for running cargo in wsl you should really install the toolchain for linux

try

```
rustup install stable-x86_64-unknown-linux-gnu --force-non-host
cargo +stable-x86_64-unknown-linux-gnu
```

but, this should work in wsl anyways, I'll try myself and see if there's any weird assumptions made that prohibits this

douweschulte:
> for running cargo in wsl you should really install the toolchain for linux
> 
> try
> 
> ```
> rustup install stable-x86_64-unknown-linux-gnu --force-non-host
> cargo +stable-x86_64-unknown-linux-gnu
> ```
> 
> but, this should work in wsl anyways, I'll try myself and see if there's any weird assumptions made that prohibits this

I am sorry to say it gives exactly the same error :/. Maybe something has gone wrong with my wsl, I have never really used it on a daily basis.

Emilgardis:
```diff
diff --git a/cargo-public-api/tests/expected-output/rustdoc_json_list.txt b/cargo-public-api/tests/expected-output/rustdoc_json_list.txt
index c55b3ea..57d257d 100644
--- a/cargo-public-api/tests/expected-output/rustdoc_json_list.txt
+++ b/cargo-public-api/tests/expected-output/rustdoc_json_list.txt
@@ -1,3 +1,6 @@
+pub const fn rustdoc_json::BuildOptions::all_features(self, all_features: bool) -> Self
+pub const fn rustdoc_json::BuildOptions::no_default_features(self, no_default_features: bool) -> Self
+pub const fn rustdoc_json::BuildOptions::quiet(self, quiet: bool) -> Self
 pub enum rustdoc_json::BuildError
 pub enum variant rustdoc_json::BuildError::CargoMetadataError(cargo_metadata::Error)
 pub enum variant rustdoc_json::BuildError::CargoTomlError(cargo_toml::Error)
@@ -10,14 +13,11 @@ pub fn rustdoc_json::BuildError::from(source: cargo_metadata::Error) -> Self
 pub fn rustdoc_json::BuildError::from(source: cargo_toml::Error) -> Self
 pub fn rustdoc_json::BuildError::from(source: std::io::Error) -> Self
 pub fn rustdoc_json::BuildError::source(&self) -> std::option::Option<&(dyn std::error::Error + 'static)>
-pub fn rustdoc_json::BuildOptions::all_features(self, all_features: bool) -> Self
 pub fn rustdoc_json::BuildOptions::default() -> Self
 pub fn rustdoc_json::BuildOptions::features<I: IntoIterator<Item = S>, S: AsRef<str>>(self, features: I) -> Self
 pub fn rustdoc_json::BuildOptions::fmt(&self, f: &mut $crate::fmt::Formatter<'_>) -> $crate::fmt::Result
 pub fn rustdoc_json::BuildOptions::manifest_path(self, manifest_path: impl AsRef<Path>) -> Self
-pub fn rustdoc_json::BuildOptions::no_default_features(self, no_default_features: bool) -> Self
 pub fn rustdoc_json::BuildOptions::package(self, package: impl AsRef<str>) -> Self
-pub fn rustdoc_json::BuildOptions::quiet(self, quiet: bool) -> Self
 pub fn rustdoc_json::BuildOptions::target(self, target: String) -> Self
 pub fn rustdoc_json::BuildOptions::toolchain(self, toolchain: impl Into<Option<String>>) -> Self
 pub fn rustdoc_json::build(options: BuildOptions) -> Result<PathBuf, BuildError>
```

apply this with `git apply diff.txt`

Enselic:
Oh, I didn't realize it was tricky to run the scripts on Windows. I ran the scripts for you now :) Feel free to merge after sanity-checking that I did not screw up. (Probably best with squash merge)

PR #156, opened 2022-09-21
Always fully qualify paths to structs, enums, unions, etc when possible

### LINKED ISSUE #145: Privately renamed items can be visible in the public API listing

Enselic:
Closes #145

### DIFF
--- a/public-api/src/intermediate_public_item.rs
+++ b/public-api/src/intermediate_public_item.rs
@@ -37,11 +37,11 @@ impl<'a> IntermediatePublicItem<'a> {
         }
     }
 
-    pub fn name(&'a self) -> String {
+    pub fn name(&self) -> &str {
         self.overridden_name
-            .clone()
-            .or_else(|| self.item.name.clone())
-            .unwrap_or_else(|| String::from("<<no_name>>"))
+            .as_deref()
+            .or(self.item.name.as_deref())
+            .unwrap_or("<<no_name>>")
     }
 
     #[must_use]
--- a/public-api/src/intermediate_public_item.rs
+++ b/public-api/src/intermediate_public_item.rs
@@ -15,7 +15,7 @@ pub struct IntermediatePublicItem<'a> {
 
     /// If `Some`, this overrides [Item::name], which happens in the case of
     /// renamed imports (`pub use other::Item as Foo;`).
-    overridden_name: Option<String>,
+    pub overridden_name: Option<String>,
 
     /// The parent item. If [Self::item] is e.g. an enum variant, then the
     /// parent is an enum. We follow the chain of parents to be able to know the
@@ -64,6 +64,11 @@ impl<'a> IntermediatePublicItem<'a> {
     #[must_use]
     pub fn path_contains_id(&self, id: &'a Id) -> bool {
         self.path().iter().any(|m| m.item.id == *id)
+    }
+
+    #[must_use]
+    pub fn path_contains_renamed_item(&self) -> bool {
+        self.path().iter().any(|m| m.overridden_name.is_some())
     }
 
     pub fn render_token_stream(&self, context: &RenderingContext) -> Vec<Token> {
--- a/public-api/src/intermediate_public_item.rs
+++ b/public-api/src/intermediate_public_item.rs
@@ -13,9 +13,9 @@ pub struct IntermediatePublicItem<'a> {
     /// The item we are effectively wrapping.
     pub item: &'a Item,
 
-    /// The name of the item. Normally this is [Item::name]. But in the case of
-    /// renamed imports (`pub use other::item as foo;`) it is the new name.
-    pub name: String,
+    /// If `Some`, this overrides [Item::name], which happens in the case of
+    /// renamed imports (`pub use other::Item as Foo;`).
+    overridden_name: Option<String>,
 
     /// The parent item. If [Self::item] is e.g. an enum variant, then the
     /// parent is an enum. We follow the chain of parents to be able to know the
@@ -27,10 +27,21 @@ impl<'a> IntermediatePublicItem<'a> {
     #[must_use]
     pub fn new(
         item: &'a Item,
-        name: String,
+        overridden_name: Option<String>,
         parent: Option<Rc<IntermediatePublicItem<'a>>>,
     ) -> Self {
-        Self { item, name, parent }
+        Self {
+            item,
+            overridden_name,
+            parent,
+        }
+    }
+
+    pub fn name(&'a self) -> String {
+        self.overridden_name
+            .clone()
+            .or_else(|| self.item.name.clone())
+            .unwrap_or_else(|| String::from("<<no_name>>"))
     }
 
     #[must_use]
--- a/public-api/src/render.rs
+++ b/public-api/src/render.rs
@@ -822,34 +822,31 @@ impl<'a> RenderingContext<'a> {
     fn best_item_for_id(&self, id: &Id) -> Option<Rc<IntermediatePublicItem<'a>>> {
         match self.id_to_items.get(&id) {
             None => None,
-            Some(items) if items.is_empty() => None,
             Some(items) => {
-                let mut candidates = items.clone();
-                candidates.sort_by(|a, b| {
-                    // If there is any item in the path that has been
-                    // renamed/re-exported, i.e. that is not the original path,
-                    // prefer that less than item with a path where all items
-                    // are original.
-                    let mut ordering = match (
-                        a.path_contains_renamed_item(),
-                        b.path_contains_renamed_item(),
-                    ) {
-                        (true, false) => Ordering::Less,
-                        (false, true) => Ordering::Greater,
-                        _ => Ordering::Equal,
-                    };
+                items
+                    .iter()
+                    .max_by(|a, b| {
+                        // If there is any item in the path that has been
+                        // renamed/re-exported, i.e. that is not the original
+                        // path, prefer that less than an item with a path where
+                        // all items are original.
+                        let mut ordering = match (
+                            a.path_contains_renamed_item(),
+                            b.path_contains_renamed_item(),
+                        ) {
+                            (true, false) => Ordering::Less,
+                            (false, true) => Ordering::Greater,
+                            _ => Ordering::Equal,
+                        };
 
-                    // If we still can't make up our mind, go with the shortest path
-                    if ordering == Ordering::Equal {
-                        ordering = b.path().len().cmp(&a.path().len());
-                    }
+                        // If we still can't make up our mind, go with the shortest path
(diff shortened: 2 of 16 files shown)


public-api/src/render.rs
```diff
@@ -17,8 +17,16 @@ macro_rules! ws {
...
 
+/// When we render an item, it might contain references to other parts of the
+/// public API. For such cases, the rendering code can use the fields in this
+/// struct.
 pub struct RenderingContext<'a> {
+    /// The original (deserialized) rustdoc JSON.
     pub crate_: &'a Crate,
+
+    /// Given a rustdoc JSON ID, keeps track of what public items that have this
+    /// ID. See [`ItemIterator::id_to_items`] for more info.
```
Emilgardis: this is technically not checked unless `document-private-items` is used, but this should be
```suggestion
    /// ID. See [`crate::item_iterator::ItemIterator::id_to_items`] for more info.
```

public-api/src/render.rs
```diff
@@ -808,6 +818,41 @@ impl<'a> RenderingContext<'a> {
...
+    fn best_item_for_id(&self, id: &Id) -> Option<Rc<IntermediatePublicItem<'a>>> {
+        match self.id_to_items.get(&id) {
+            None => None,
+            Some(items) if items.is_empty() => None,
+            Some(items) => {
+                let mut candidates = items.clone();
+                candidates.sort_by(|a, b| {
+                    // If there is any item in the path that has been
+                    // renamed/re-exported, i.e. that is not the original path,
+                    // prefer that less than item with a path where all items
```
Emilgardis: typo?

```
prefer that less than item
```

public-api/src/intermediate_public_item.rs
```diff
@@ -27,10 +27,21 @@ impl<'a> IntermediatePublicItem<'a> {
...
+            parent,
+        }
+    }
+
+    pub fn name(&'a self) -> String {
+        self.overridden_name
+            .clone()
+            .or_else(|| self.item.name.clone())
+            .unwrap_or_else(|| String::from("<<no_name>>"))
     }
```
Emilgardis: this doesn't need to clone, I personally think it's better to clone/to_owned on call site when feasible.

```suggestion
    pub fn name(&self) -> &str {
        self.overridden_name.as_deref()
            .or(self.item.name.as_deref())
            .unwrap_or("<<no_name>>")
    }
```

not also that the `fn name(&'a self)` is overly restrictive over the lifetime, it's not needed. Same goes for the `fn path(&'a self)` below

public-api/src/render.rs
```diff
@@ -808,6 +818,41 @@ impl<'a> RenderingContext<'a> {
...
+                        ordering = b.path().len().cmp(&a.path().len());
+                    }
+
+                    ordering
+                });
+
+                // Note that we `pop()` here, so in the sorting above we want to
+                // the best element to end up last.
+                candidates.pop()
+            }
```
Emilgardis: this could use `vec.iter().max_by(|a,b| ...).cloned()`

```suggestion
           Some(items) => {
                items
                    .iter()
                    .max_by(|a, b| {
                        // If there is any item in the path that has been
                        // renamed/re-exported, i.e. that is not the original path,
                        // prefer that less than item with a path where all items
                        // are original.
                        let mut ordering = match (
                            a.path_contains_renamed_item(),
                            b.path_contains_renamed_item(),
                        ) {
                            (true, false) => Ordering::Less,
                            (false, true) => Ordering::Greater,
                            _ => Ordering::Equal,
                        };

                        // If we still can't make up our mind, go with the shortest path
                        if ordering == Ordering::Equal {
                            ordering = b.path().len().cmp(&a.path().len());
                        }

                        ordering
                    })
                    .cloned()
            }
```

Emilgardis: this is also a good opportunity for a new clippy lint maybe

Emilgardis (approved):
looks good, some nits!

public-api/src/intermediate_public_item.rs
Enselic: (Thanks for this and other review comments!) Agreed, I tried to make this return &'a str at first, but couldn't make the borrow checker happy, so I gave up. I will try again, maybe me using `'a` was the problem.

public-api/src/render.rs
Enselic: Very nice 👍 

I think I took care of this and the other comments now.

test-apis/comprehensive_api/src/exports.rs
```diff
@@ -26,3 +26,38 @@ pub mod recursion_glob_1 {
...
+    pub fn privately_renamed_external_arg_type(_transform: privately_renamed::External) {}
+
+    pub mod external_2 {
+        pub struct External;
+    }
+
+    pub use external_2 as publicly_renamed;
+    use external_2::External as PrivatelyRenamed2;
+
+    pub fn privately_renamed_2_arg_type(_transform: PrivatelyRenamed2) {} // HTML difference!
```
Emilgardis: is this an issue?

Enselic: Since rustc/rustdoc accepts this code, we can assume that there is real code out there that does this, but I agree it is a bit silly. But since presumably this code exists in the while, I think it makes sense to test for it. If this ends up being a lot of work to maintain, we can give up and remove the test.

Emilgardis: It could be easy to miss that this shouldn't change when blessing test output, perhaps a check for no leaked private idents/paths is needed?

Enselic: Sorry, I am on a small laptop, so I didn't see that you commented on `// HTML difference`!  Cause that's what you mean, right? That is a comment I accidentally left in. It was just some note keeping I used when verifying my change. The current output is the same as HTML. I will remove this obsoleted comment.

test-apis/comprehensive_api/src/exports.rs
```diff
@@ -26,3 +26,38 @@ pub mod recursion_glob_1 {
 pub mod recursion_glob_2 {
     pub use super::recursion_glob_1::*;
 }
+
+/// Regression test for <https://github.com/Enselic/cargo-public-api/issues/145>
+pub mod issue_145 {
```
Emilgardis, line 31: It could be easy to miss that this shouldn't change when blessing test output, perhaps a check for no leaked private idents/paths is needed?

Commented [here](https://github.com/Enselic/cargo-public-api/pull/156#discussion_r978936784)

test-apis/comprehensive_api/src/exports.rs
Enselic: > It could be easy to miss that this shouldn't change when blessing test output, perhaps a check for no leaked private idents/paths is needed?

Hmm good point. Probably quite messy to add such a check though. Let's try to be careful :P

test-apis/comprehensive_api/src/exports.rs
Enselic: Good point. Probably quite messy to add such a check though. Let's try to be careful :P

Emilgardis: I think what could be done is put some sequence of characters on the idents that should never be present in the output, and then check for it.

So maybe check that there is nothing matching `private_noleak`/`PrivateNoLeak` or something

Enselic: Yep that is also how I would do it, and I agree it should be pretty solid. But it is not foolproof, because when new code is added, one has to be careful to follow that naming pattern. I guess we could add a check for that too though. But, messy :)

Emilgardis: wouldn't the check for no existence of that sequence be that check? ;)

Enselic: I might very well be missing something, but `PrivateButNamedInTheWrongWay` would inappropriately pass that check, wouldn't it?

Emilgardis: but that's just a wrong test case, no?

e.g adding a test to check that the test is correct sounds needless, as that should be catched easily in review

Enselic: Well, I wouldn't trust myself to remember to think about such things in review :P

Emilgardis: I think this could be considered for later, the pr is good with and without this imo.

feel free to squash the commits, no need for attribute me either

PR #158, opened 2022-09-21
move `rustdoc_json::build` into a struct fn `BuildOptions::build`

Emilgardis:
also deprecates `rustdoc_json::build` for this new method

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -11,7 +11,7 @@ use public_api::diff::PublicItemsDiff;
 use public_api::{Options, PublicApi, PublicItem, MINIMUM_RUSTDOC_JSON_VERSION};
 
 use clap::Parser;
-use rustdoc_json::{BuildError, BuildOptions};
+use rustdoc_json::BuildError;
 
 mod arg_types;
 mod error;
@@ -365,21 +365,21 @@ fn collect_public_api_from_commit(
     } else {
         None
     };
-    let mut build_options = BuildOptions::default()
+    let mut builder = rustdoc_json::Builder::default()
         .toolchain(args.toolchain.clone())
         .manifest_path(&args.manifest_path)
         .all_features(args.all_features)
         .no_default_features(args.no_default_features)
         .features(&args.features);
     if let Some(target) = &args.target {
-        build_options = build_options.target(target.clone());
+        builder = builder.target(target.clone());
     }
 
     if let Some(package) = &args.package {
-        build_options = build_options.package(package);
+        builder = builder.package(package);
     }
 
-    let json_path = match rustdoc_json::build(build_options) {
+    let json_path = match builder.build() {
         Err(BuildError::VirtualManifest(manifest_path)) => virtual_manifest_error(&manifest_path)?,
         res => res?,
     };
--- a/public-api/examples/diff_public_api.rs
+++ b/public-api/examples/diff_public_api.rs
@@ -1,23 +1,20 @@
 use std::{error::Error, fs::read_to_string};
 
 use public_api::{diff::PublicItemsDiff, Options, PublicApi};
-use rustdoc_json::BuildOptions;
 
 fn main() -> Result<(), Box<dyn Error>> {
     let options = Options::default();
 
-    let old_json = rustdoc_json::build(
-        BuildOptions::default()
-            .toolchain(String::from("+nightly"))
-            .manifest_path("test-apis/example_api-v0.1.0/Cargo.toml"),
-    )?;
+    let old_json = rustdoc_json::Builder::default()
+        .toolchain(String::from("+nightly"))
+        .manifest_path("test-apis/example_api-v0.1.0/Cargo.toml")
+        .build()?;
     let old = PublicApi::from_rustdoc_json_str(&read_to_string(old_json)?, options)?;
 
-    let new_json = rustdoc_json::build(
-        BuildOptions::default()
-            .toolchain(String::from("+nightly"))
-            .manifest_path("test-apis/example_api-v0.2.0/Cargo.toml"),
-    )?;
+    let new_json = rustdoc_json::Builder::default()
+        .toolchain(String::from("+nightly"))
+        .manifest_path("test-apis/example_api-v0.2.0/Cargo.toml")
+        .build()?;
     let new = PublicApi::from_rustdoc_json_str(&read_to_string(new_json)?, options)?;
 
     let diff = PublicItemsDiff::between(old.items, new.items);
--- a/public-api/examples/list_public_api.rs
+++ b/public-api/examples/list_public_api.rs
@@ -2,14 +2,11 @@ use std::{error::Error, fs::read_to_string};
 
 use public_api::{Options, PublicApi};
 
-use rustdoc_json::BuildOptions;
-
 fn main() -> Result<(), Box<dyn Error>> {
-    let json_path = rustdoc_json::build(
-        BuildOptions::default()
-            .toolchain(String::from("+nightly"))
-            .manifest_path("test-apis/example_api-v0.2.0/Cargo.toml"),
-    )?;
+    let json_path = rustdoc_json::Builder::default()
+        .toolchain(String::from("+nightly"))
+        .manifest_path("test-apis/example_api-v0.2.0/Cargo.toml")
+        .build()?;
 
     let public_api =
         PublicApi::from_rustdoc_json_str(&read_to_string(&json_path)?, Options::default())?;
--- a/rustdoc-json/CHANGELOG.md
+++ b/rustdoc-json/CHANGELOG.md
@@ -1,3 +1,8 @@
+## v0.4.1
+* rename `BuildOptions` to `Builder`, a new insta-deprecated alias `BuildOptions` is available
+* deprecate `build()`, replaced with `Builder::build()`
+* make `BuildOptions`, replaced with `Builder::build()`
+
 ## v0.4.0
 * Support for specifying `--target`, `--features`, and `--package`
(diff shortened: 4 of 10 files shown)


rustdoc-json/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## v0.4.1
+* deprecate `build()`, replaced with `BuildOptions::build()`
+
```
Emilgardis: not sure how the changelog is handled, should this maybe be 

```suggestion
## Unreleased
* deprecate `build()`, replaced with `BuildOptions::build()`

```

?

Enselic (commented):
Hmm, I would expect `BuildOptions::build()` to produce a `BuildOptions`. Maybe we should introduce a `struct RustdocJson` with `fn path(&self) -> &Path` and `fn build(options: BuildOptions) -> Self`? Unsure if it is worth the extra code though, until we need to return more than one field, so to speak.

rustdoc-json/CHANGELOG.md
Enselic: This changelog is quite un-formal, I just wanted to have something rather than nothing. But I am totally fine with deciding on the next version directly. In fact, I prefer it, because otherwise someone else has to figure out the version later.

Emilgardis:
I agree, the naming is a bit awkward

I'd say `BuildOptions` is well suited to be a builder. This is a [common pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) found for example in [`hyper`](https://docs.rs/hyper/0.14.20/hyper/client/struct.Builder.html#method.build) and other crates. 

I think maybe `BuildOptions` could be renamed to `Builder` or maybe simply `Build`? and we could either return a `Result<PathBuf, _>` or a new struct `Result<RustdocJson, _>`

the rename can be done pretty easily with a alias, `#[deprecated]` on `pub use Build as BuildOptions;` _should_ only deprecates the alias, see for example how [`clap`](https://github.com/clap-rs/clap/blob/7de2f36e6b82efafa0b3575f871ed2f262e123bc/src/lib.rs#L143-L146) did it.

Emilgardis:
Maybe even `RustdocJsonBuilder`?

Enselic:
I think you're onto something here.

What do you think about this? (I think I like it)

```rust
    let mut json_path = rustdoc_json::Builder::default()
        .toolchain(args.toolchain.clone())
        .manifest_path(&args.manifest_path)
        .build()
```

So like `RustdocJsonBuilder`, but we keep the crate name in the code so that the code reads nicely (similar to how I went with `rustdoc_json::build()` instead of `rustdoc_json::rustdoc_json_build()`).

public-api/examples/diff_public_api.rs
```diff
@@ -1,23 +1,21 @@
...
 
 fn main() -> Result<(), Box<dyn Error>> {
     let options = Options::default();
 
-    let old_json = rustdoc_json::build(
-        BuildOptions::default()
-            .toolchain(String::from("+nightly"))
-            .manifest_path("test-apis/example_api-v0.1.0/Cargo.toml"),
-    )?;
+    let old_json = Builder::default()
```
Enselic: (`rustdoc_json::Builder` instead?)

cargo-public-api/src/main.rs
```diff
@@ -365,21 +365,21 @@ fn collect_public_api_from_commit(
     } else {
         None
     };
-    let mut build_options = BuildOptions::default()
+    let mut builder = Builder::default()
```
Enselic: Wouldn't it be better to make it clear what kind of builder this is directly at this first-use site?
```suggestion
    let mut builder = rustdoc_json::Builder::default()
```
and same thing for the other use-sites below.

public-api/examples/diff_public_api.rs
```diff
@@ -1,23 +1,21 @@
...
+        .manifest_path("test-apis/example_api-v0.1.0/Cargo.toml")
+        .build()?;
     let old = PublicApi::from_rustdoc_json_str(&read_to_string(old_json)?, options)?;
 
-    let new_json = rustdoc_json::build(
-        BuildOptions::default()
-            .toolchain(String::from("+nightly"))
-            .manifest_path("test-apis/example_api-v0.2.0/Cargo.toml"),
-    )?;
+    let new_json = Builder::default()
```
Enselic: (`rustdoc_json::Builder` instead?)

public-api/examples/list_public_api.rs
```diff
@@ -2,14 +2,13 @@ use std::{error::Error, fs::read_to_string};
...
-use rustdoc_json::BuildOptions;
+use rustdoc_json::Builder;
 
 fn main() -> Result<(), Box<dyn Error>> {
-    let json_path = rustdoc_json::build(
-        BuildOptions::default()
-            .toolchain(String::from("+nightly"))
-            .manifest_path("test-apis/example_api-v0.2.0/Cargo.toml"),
-    )?;
+    let json_path = Builder::default()
```
Enselic: (`rustdoc_json::Builder` instead?)

rustdoc-json/examples/build-rustdoc-json.rs
```diff
@@ -3,14 +3,13 @@
...
-    use rustdoc_json::BuildOptions;
+    use rustdoc_json::Builder;
 
     // Build it
-    let json_path = rustdoc_json::build(
-        BuildOptions::default()
-            .toolchain("+nightly".to_owned())
-            .manifest_path(&std::env::args().nth(1).unwrap()),
-    )?;
+    let json_path = Builder::default()
```
Enselic: (`rustdoc_json::Builder` instead?)

rustdoc-json/src/lib.rs
```diff
@@ -2,17 +2,17 @@
...
 //! ```
-//!    use rustdoc_json::BuildOptions;
+//! use rustdoc_json::Builder;
 //!
-//!    let json_path = rustdoc_json::build(
-//!        BuildOptions::default()
-//!            .toolchain("+nightly".to_owned())
-//!            .manifest_path("Cargo.toml"),
-//!    ).unwrap();
+//! let json_path = Builder::default()
```
Enselic: (`rustdoc_json::Builder` instead?)

rustdoc-json/src/lib.rs
```diff
@@ -25,7 +25,10 @@ use std::path::PathBuf;
 
 mod build;
 
-/// Represents all errors that can occur when using [`crate::build()`].
+#[deprecated(note = "this struct has been renamed to `rustdoc_json::Builder`")]
+pub use Builder as BuildOptions;
```
Enselic, line 28: Clever trick 👍 

I think we can be pretty liberal in removing deprecated APIs, so IMHO we don't have to wait long to remove the deprecated API, maybe just one or two 0.x.0 bumps, if at all.

I still consider these APIs to be quite experimental, and we don't have that many external clients.

test-utils/src/lib.rs
```diff
@@ -17,13 +17,13 @@ pub fn rustdoc_json_path_for_crate(test_crate: &str) -> PathBuf {
...
     // so build quietly to make running tests much less noisy
-    rustdoc_json::build(
-        BuildOptions::default()
-            .toolchain("+nightly".to_owned())
-            .manifest_path(&format!("{}/Cargo.toml", test_crate))
-            .quiet(true),
-    )
-    .unwrap()
+
+    Builder::default()
```
Enselic: (`rustdoc_json::Builder` instead?)

Enselic (approved):
Nice, looks good! Just one question/comment regarding if not `rustdoc_json::Builder` improves readability of the code. IMHO it does, but I will let you have the final say. Feel free to merge this once you have made up your mind.

Enselic:
(Needs to be rebased on https://github.com/Enselic/cargo-public-api/pull/156, otherwise CI will fail after merge)

Emilgardis:
I like the idea of using the full path, will fix

Enselic (approved):
🙏

Emilgardis:
should we do an immediate bump/release also?

Enselic:
Do you mean of only `rustdoc_json`? Maybe wait a few days in case we discover we want to do some tweaks?

But if you want to release right away, you have my full endorsement :)

Emilgardis:
Precisely, I think it's fine to wait then :)

merging this then 🚀

rustdoc-json/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## v0.4.1
+* rename `BuildOptions` to `Builder`, a new insta-deprecated alias `BuildOptions` is available
+* deprecate `build()`, replaced with `Builder::build()`
+* make `BuildOptions`, replaced with `Builder::build()`
```
Emilgardis, line 4: oops, this is redundant

Enselic: feel free to push such tweaks directly to main without going through a PR (as also mentioned in MAINTAINER.md)

Enselic:
@Emilgardis: Tomorrow's nightly changes the JSON format slightly, so we're gonna need to make a release: https://github.com/rust-lang/rust/pull/102321#event-7464506844

Are you interested in taking care of this release in full, or only the `rustdoc_json` crate?

Enselic:
Turns out we have no test that exercises the `ItemEnum::PrimitiveType` code path, so CI still passes.

Actually we might get away with not bumping min nightly, because only `core` can `impl` stuff for primities.

Enselic:
FYI, rustdoc-json v0.4.1 has now been published to crates.io.

Emilgardis:
my bad! I missed the earlier comment!

Enselic:
No worries :)

(I'm doing the v0.20.0 release now, mostly because I am not comfortable with accumulating unreleased stuff)

PR #160, opened 2022-09-21
more docs

Emilgardis:


Emilgardis:
@SUPERCILEX do you think this would be helpful for you?

SUPERCILEX:
Oh snap, you can run the JSON gen command within the test? Yes, that's super useful, thanks!

Enselic (approved):
Nice! (But needs to be updated to match the other approved PR of course)

PR #162, opened 2022-09-23
CI: Enforce `--document-private-items` passes without warnings

Enselic:
So that e.g. https://github.com/Enselic/cargo-public-api/pull/156#discussion_r978820594 is automatically detected and everyone saves some time :)


Emilgardis (approved):
why not :D

PR #165, opened 2022-09-25
Replace most of `bless-expected-output-for-tests.sh` with `BLESS=1` trick

Enselic:
This makes it faster to bless output since it happens in parallel. On my machine it now takes ~10 seconds to bless instead of ~15 seconds. Can probably be improved by extending the `BLESS=1` trick to even more tests.

Kudos to Emilgardis for teaching me about the `BLESS=1` trick.


Emilgardis (approved):
looks good!

could also consider switching to [`snapbox`](https://docs.rs/snapbox/0.4.0/snapbox/) or [`trycmd`](https://docs.rs/trycmd/latest/trycmd/) in the future which both have this concept of writing/blessing output

feel free to merge

Enselic:
Thank you for the review and the tips, I never heard of them. Will be my go-to if we run into a dead-end with this solution or if it becomes too much work to maintain.

PR #168, opened 2022-10-01
Make `--toolchain` take toolchain name without `+`

### LINKED ISSUE #161: specify toolchain without + on `--toolchain` and co.
https://github.com/Enselic/cargo-public-api/blob/6c84ec97a2c4d7e8137bd872bf6fe39906e3855f/rustdoc-json/src/build.rs#L54

----

Maybe it would make more sense to change to an interface where toolchains are specified without +, so that we don't have to trim + here?

_Originally posted by @Enselic in https://github.com/Enselic/cargo-public-api/pull/147#discussion_r971517104_

Enselic:
A Rust "toolchain" is actually referred to without a `+`. The `+` comes from the proxy mechanism of rustup, but `+` is not actually part of the toolchain name.

See for example https://rust-lang.github.io/rustup/concepts/toolchains.html:

```
rustup toolchain install stable-x86_64-pc-windows-msvc
```

There is no `+` there.

Closes #161

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -102,13 +102,13 @@ pub struct Args {
     #[clap(long, hide = true)]
     verbose: bool,
 
-    /// Allows you to build rustdoc JSON with a toolchain other than `+nightly`.
+    /// Allows you to build rustdoc JSON with a toolchain other than `nightly`.
     ///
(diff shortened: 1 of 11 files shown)


Emilgardis (approved):
lgtm!

PR #171, opened 2022-10-02
Allow to specify `--target-dir` when building rustdoc JSON

Enselic:
Mostly to fix flaky tests caused by rustdoc JSON buliding race conditions, reproducible for me with:

    while true; do cargo clean -p public-api ; cargo test -p public-api; sleep 1; done 2>&1 | grep 'test result'

I got a test failure approximately 2% of the time before the fix.

### DIFF
--- a/public-api/tests/public-api-bin-tests.rs
+++ b/public-api/tests/public-api-bin-tests.rs
@@ -202,8 +202,8 @@ fn cmd_with_rustdoc_json_args(crates: &[&str], final_steps: impl FnOnce(Command)
     let mut temp_dirs = vec![];
 
     for crate_ in crates {
-        // Put output in a temp dir so that concurrently running tests do not
-        // concurrently write and read the same JSON, which causes tests to fail
+        // Put output in a temp dir to ensure concurrently running tests do not
+        // write and read the same JSON simultaneously, which causes tests to fail
         // sporadically.
         let temp_dir = TempDir::new().unwrap();
 
--- a/public-api/tests/public-api-bin-tests.rs
+++ b/public-api/tests/public-api-bin-tests.rs
@@ -9,8 +9,10 @@ use public_api::MINIMUM_RUSTDOC_JSON_VERSION;
 // rust-analyzer bug: https://github.com/rust-lang/rust-analyzer/issues/9173
 #[path = "../../test-utils/src/lib.rs"]
 mod test_utils;
+use tempfile::TempDir;
 use test_utils::assert_or_bless::AssertOrBless;
 use test_utils::rustdoc_json_path_for_crate;
+use test_utils::rustdoc_json_path_for_crate_with_target_dir;
 
 #[test]
 fn print_public_api() {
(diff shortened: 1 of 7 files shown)


rustdoc-json/src/build.rs
```diff
@@ -176,6 +187,15 @@ impl Builder {
         self
     }
 
+    /// Set what `--target-dir` to pass to `cargo`. Typically only needed if you
+    /// want to be able to build rustdoc JSON for the same crate concurerntly,
+    /// for example to paralellize regression tests.
```
Emilgardis:
```suggestion
    /// for example to parallelize regression tests.
```

rustdoc-json/src/build.rs
```diff
@@ -176,6 +187,15 @@ impl Builder {
         self
     }
 
+    /// Set what `--target-dir` to pass to `cargo`. Typically only needed if you
+    /// want to be able to build rustdoc JSON for the same crate concurerntly,
```
Emilgardis:
```suggestion
    /// want to be able to build rustdoc JSON for the same crate concurrently,
```

public-api/tests/public-api-bin-tests.rs
```diff
@@ -197,9 +199,29 @@ To include blanket implementations, pass --with-blanket-implementations.
 fn cmd_with_rustdoc_json_args(crates: &[&str], final_steps: impl FnOnce(Command)) {
     let mut cmd = Command::cargo_bin("public-api").unwrap();
 
+    let mut temp_dirs = vec![];
+
     for crate_ in crates {
-        cmd.arg(rustdoc_json_path_for_crate(crate_));
+        // Put output in a temp dir so that concurrently running tests do not
+        // concurrently write and read the same JSON, which causes tests to fail
+        // sporadically.
```
Emilgardis:
```suggestion
        // Put output in a temp dir to ensure concurrently running tests do not
        // write and read the same JSON simultaneously, which causes tests to fail
        // sporadically.
```

public-api/tests/public-api-bin-tests.rs
```diff
@@ -197,9 +199,29 @@ To include blanket implementations, pass --with-blanket-implementations.
...
+        // is used, so using the same dir for all would cause JSON files to be
+        // overwitten.
+        temp_dirs.push(temp_dir);
     }
 
     final_steps(cmd);
+
+    // Here temp_dirs are dropped/removed. To prevent that for debugging
+    // purposes, wrap `TempDir::new().unwrap()` in
+    // `std::mem::ManuallyDrop::new()`, but never drop.
```
Emilgardis: This comment is a bit over-complicated.

Instead, [`TempDir::into_path`](https://docs.rs/tempfile/3.3.0/tempfile/struct.TempDir.html#method.into_path) should be used

Emilgardis (changes requested):
Nice!

This could be added as an option to `cargo public-api` too, e.g `cargo public-api --target-dir foo`

rustdoc-json/src/build.rs
Enselic: Oops! Normally I have a spell-checker installed to catch typos like these, but on my current system I had forgot to install it. Will fix :)

public-api/tests/public-api-bin-tests.rs
Enselic: Aha! Thanks for pointing that out :pray:

Enselic:
Thanks for review! All comments should have been taken care of. (Probably best to merge this as a squash merge since I did a bunch of CR-fixup commits.)

> This could be added as an option to `cargo public-api` too, e.g `cargo public-api --target-dir foo`

I agree, although not sure that has any use for regular users, or even us. Whoever needs it first gets the honor to implement it :D

Emilgardis (approved):
lgtm

PR #174, opened 2022-10-07
Add option `--force-git-checkouts` to force git checkouts

### LINKED ISSUE #172: Add option to force git checkouts
I see from the `--diff-git-checkouts` option description that inability to do `git checkout` on a dirty tree is a feature. However, there are cases where it's not desirable. In our case rust build script produces some artefacts that are supposed to be checked in the repo. So, when the tool checks out the first commit and generates the docs the tree is not clean and the tools fails to checkout the second commit.

Would you accept a PR implementing an additional `--force-git-checkouts` option?

inikulin:
Closes #172

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -70,6 +70,11 @@ pub struct Args {
     #[clap(long, min_values = 2, max_values = 2)]
     diff_git_checkouts: Option<Vec<String>>,
 
+    /// Raise this flag to discard working tree changes during git checkouts when
+    /// `--diff-git-checkouts` is used.
+    #[clap(long)]
+    force_git_checkouts: bool,
+
     /// Usage: --diff-rustdoc-json <RUSTDOC_JSON_PATH_1> <RUSTDOC_JSON_PATH_2>
     ///
     /// Diff the public API across two different rustdoc JSON files.
@@ -318,7 +323,12 @@ fn print_diff(args: &Args, old: Vec<PublicItem>, new: Vec<PublicItem>) -> Result
 impl PostProcessing {
     fn perform(&self, args: &Args) -> Result<()> {
         if let Some(branch_to_restore) = &self.branch_to_restore {
-            git_utils::git_checkout(branch_to_restore, &args.git_root()?, !args.verbose)?;
+            git_utils::git_checkout(
+                branch_to_restore,
+                &args.git_root()?,
+                !args.verbose,
+                args.force_git_checkouts,
+            )?;
         }
 
         check_diff(args, &self.diff_to_check)
@@ -366,6 +376,7 @@ fn collect_public_api_from_commit(
             commit,
             &args.git_root()?,
             !args.verbose,
+            args.force_git_checkouts,
         )?)
     } else {
         None
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -26,6 +26,24 @@ use test_utils::rustdoc_json_path_for_crate;
 #[path = "../src/git_utils.rs"] // Say NO to copy-paste!
 mod git_utils;
 
+fn create_test_repo_with_dirty_git_tree() -> TestRepo {
+    let test_repo = TestRepo::new();
+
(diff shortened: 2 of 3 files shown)


cargo-public-api/src/main.rs
```diff
@@ -70,6 +70,11 @@ pub struct Args {
     #[clap(long, min_values = 2, max_values = 2)]
     diff_git_checkouts: Option<Vec<String>>,
 
+    /// Raise this flag to discard working tree changes during git checkouts when
+    /// `--diff-git-checkouts` is used.
+    #[clap(long)]
+    force_git_checkouts: bool,
```
Enselic, line 76: Maybe shorten this to just `--force`? So that we can use it in more contexts later. Unsure if this is a good idea though. What do you think?

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -201,6 +201,38 @@ fn diff_public_items_with_dirty_tree_fails() {
...
+
+    let mut lib_rs = OpenOptions::new()
+        .write(true)
+        .append(true)
+        .open(&lib_rs_path)
+        .unwrap();
+
+    writeln!(lib_rs, "// Make git tree dirty").unwrap();
+
+    // Make sure diffing does not destroy uncommitted data!
```
Enselic: Comment needs to be updated (or maybe just removed)

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -201,6 +201,38 @@ fn diff_public_items_with_dirty_tree_fails() {
...
 }
 
+/// Test that diffing succeedes if the git tree is dirty and
+/// `force-git-checkout` option is specified.
+#[test]
+#[cfg_attr(target_family = "windows", ignore)]
+fn diff_public_items_with_dirty_tree_succeedes_with_force_option() {
+    let test_repo = TestRepo::new();
+
+    // Make the tree dirty by appending a comment to src/lib.rs
```
Enselic: Optional: Would be nice to add a helper for this rather than duplicating the code, but not required. That can always be done later.

Enselic (commented):
Looks great! Just had some minor comments.

cargo-public-api/src/main.rs
Emilgardis: it could be gated behind a group with the other git flags and renamed --force

https://github.com/clap-rs/clap/blob/f513862939239a713734bc80537e34bf1bbd2661/examples/tutorial_derive/04_03_relations.rs#L5

inikulin: > Maybe shorten this to just --force? So that we can use it in more contexts later. Unsure if this is a good idea though. What do you think?

I would say having explicit option that explains what's really enforced is more preferable. `--force` is too vague and can have undesirable side-effects if extended with more functionality later. 

> it could be gated behind a group with the other git flags and renamed --force

I thought about that, but that would introduce a CLI breaking change.

These are my 2c, but it's your call

cargo-public-api/tests/cargo-public-api-bin-tests.rs
inikulin: Good point, do you want me to put those `test-utils` or having helper functions in this file is fine?

cargo-public-api/src/main.rs
Enselic: I'm fine with keeping it as you have it, and then maybe re-design the CLI sometime in the future.

@Emilgardis Is that OK with you too?

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Enselic: I prefer if you keep the new helper in this file for now. We can then move it to `test-utils` when it is needed in some other place (if ever)

cargo-public-api/src/main.rs
Emilgardis: yep!

inikulin:
Addressed the comments

Enselic (approved):
LGTM, thanks!

(I will tweak the PR title a bit since it is used verbatim in the upcoming auto-generated release notes)

PR #179, opened 2022-10-12
Bump `cargo_toml` to `0.13.0-alpha.1` to not crash on `version.workspace = true`

### LINKED ISSUE #175: Toml parsing fails if it uses workspace keys
This:

```toml
[package]
version.workspace = true
```

Causes the following crash
```
thread 'api' panicked at 'called `Result::unwrap()` on an `Err` value: CargoTomlError(Parse(Error { inner: ErrorInner { kind: Custom, line: Some(2), col: 20, at: Some(50), message: "invalid type: map, expected a string", key: ["package", "version"] } }))', fuc_engine/tests/api.rs:10:82
```

Not using `*.workspace = true` fixes the issue.

Enselic:
Not crashing is more important than not depending on alphas IMHO, even if it feels a bit awkward to depend on alphas. Curious to hear what others thinks about this.

Closes #175

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -117,12 +117,11 @@ dependencies = [
 
 [[package]]
 name = "cargo_toml"
-version = "0.11.5"
+version = "0.13.0-alpha.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Enselic:
Closing for now, see https://github.com/Enselic/cargo-public-api/issues/175#issuecomment-1277022702

PR #181, opened 2022-10-14
rustdoc-json: Remove most of `BuildOptions`, only leave deprecation message

Enselic:
Our current deprecation setup is not aggressive enough, because this code compiles without warning:

https://github.com/SUPERCILEX/ftzz/blob/23c80a1c3c3372254311d9798409ad82f2564676/tests/api.rs#L9-L10

I have created a PR with a fix: https://github.com/SUPERCILEX/ftzz/pull/5

But I think it is time now to remove BuildOptions. But still keep it around for the sake of a deprecation message.

I could not figure out a way to keep the v0.4.x API while still getting a deprecation message. There is already a deprecation message, but it does not work due to https://github.com/rust-lang/rust/issues/30827. So prepare bump to v0.5.0.

### DIFF
--- a/cargo-public-api/tests/expected-output/rustdoc_json_list.txt
+++ b/cargo-public-api/tests/expected-output/rustdoc_json_list.txt
@@ -1,16 +1,18 @@
 impl !RefUnwindSafe for rustdoc_json::BuildError
 impl !UnwindSafe for rustdoc_json::BuildError
+impl RefUnwindSafe for rustdoc_json::BuildOptions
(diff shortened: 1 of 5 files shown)


Emilgardis (approved):

Emilgardis:
I'm not sure I agree with that this is a case of the deprecation not being aggressive enough, it's just a very strange usage of the library

Emilgardis:
Feel free to merge!

Enselic:
"aggressive" was probably not the best choice of word on my part 😅 

Thanks for review; merging

PR #182, opened 2022-10-14
Validate and resolve CLI commit references

### LINKED ISSUE #180: Pre-resolve git references to support e.g. `HEAD` as second git reference
Specifying `HEAD` instead of `$(git rev-parse HEAD)` should work but it only works if it's the first argument to `--diff-git-checkouts` because its "value" is lost if a different checkout was specified.

i.e.

```
cargo public-api --diff-git-checkouts 0.4.0 HEAD
```

First checks out 0.4.0 then checks out HEAD, but HEAD is now 0.4.0.

In case HEAD is specified, I believe `cargo-public-api` should first internally call `git rev-parse HEAD` to determine the equivalent commit then use that to checkout later.

mqudsi:
Validate the existence of provided commits and resolve "relative" git refs or committish values to their absolute equivalents.

Allows values like `HEAD` to be used in specifying commits to diff.

Closes #180.

### DIFF
--- a/cargo-public-api/src/git_utils.rs
+++ b/cargo-public-api/src/git_utils.rs
@@ -84,3 +84,10 @@ fn trimmed_stdout(mut cmd: Command) -> Result<String> {
         Err(anyhow!("Failure: {:?}", output))
     }
 }
+
+/// Resolves a git reference provided at the CLI to an actual commit, allowing
+/// us to validate refs and use "relative" values like HEAD and more.
+#[allow(unused)] // It IS used!
+pub fn resolve_ref(path: impl AsRef<Path>, committish: &str) -> Result<String> {
+    trimmed_git_stdout(path, &["rev-parse", committish])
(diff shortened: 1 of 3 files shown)


Enselic (commented):
Looks good on first glance!

Can you also remove these two lines:

https://github.com/Enselic/cargo-public-api/blob/8c808fc8546df3705e295cb4b028fce13483a465/cargo-public-api/src/main.rs#L63-L64

and also add a test please? I think a single test that diffs between HEAD^ and HEAD is sufficient, because such a test I think only passes with this fix. There are many git diff tests already, so should be pretty easy to add this one. Of course feel free to reach out if you get stuck.

mqudsi:
OK, I've added the tests and updated the documentation to reflect `cargo-public-api`'s newfound abilities.

Enselic (approved):
Looks good, thanks!

PR #185, opened 2022-10-16
Show fully qualified paths for external items

### LINKED ISSUE #183: Qualified path not shown for all types in signatures
I was hoping to use `cargo-public-api` to check for accidentally including dependency type in public APIs, for example:

```rust
use aws_sdk_cloudformation::model::Tag; // import from a dependency

// ...

pub fn set_tags(tags: Vec<Tag>) { /* ... */ }
```

What I had hoped to see in the output for this example (in a crate called `pubapicheck`) is:

```
pub fn pubapicheck::set_tags(tags: Vec<aws_sdk_cloudformation::model::Tag>)
pub mod pubapicheck
```

But what I actually get is:

```
pub fn pubapicheck::set_tags(tags: Vec<Tag>)
pub mod pubapicheck
```

You can narrow things down a bit with some gnarly regex shenanigans to find types without your crate (or other common words) as a prefix, but it would be easier if types in signatures were presented by their qualified path.

<details>

<summary>gnarly regex shenanigans</summary>

```sh
cargo public-api \
  | rg -P '(?<=[^_a-zA-Z0-9:'"'"'])(?!cloudformatious(::|\b)|pub|unsafe|type|struct|field|trait|String|Vec|Option|Result|mut|str|self|u\d+|Self|fn|mod|enum|variant|impl|Send|Sync|for|(Ref)?UnwindSafe|Unpin|std|bool|Into|where|async|const|Box|dyn|Pin|crate|_serde|__S|__D|as\b|task::|fmt::|Iterator\b|Item\b|D\b|S\b|_\b|I\b|non_exhaustive)([_a-zA-Z0-9]+::)*[_a-zA-Z0-9]+(?!:)\b'
```

</details>

Apologies if this is a known limitation.

Enselic:
Using the rustdoc JSON `paths` map.

Closes #183 

If we are unlucky, this will stop working with https://github.com/rust-lang/rust/pull/103085, but I have requested the change to remain compatible with this PR (I have verified that just keeping `path` unchanged in that PR keeps our code working)


rustdoc-json/public-api.txt
```diff
@@ -17,28 +17,28 @@ pub const fn rustdoc_json::Builder::all_features(self, all_features: bool) -> Se
...
 pub fn rustdoc_json::Builder::default() -> Self
-pub fn rustdoc_json::Builder::features<I: IntoIterator<Item = S>, S: AsRef<str>>(self, features: I) -> Self
-pub fn rustdoc_json::Builder::fmt(&self, f: &mut $crate::fmt::Formatter<'_>) -> $crate::fmt::Result
-pub fn rustdoc_json::Builder::manifest_path(self, manifest_path: impl AsRef<Path>) -> Self
-pub fn rustdoc_json::Builder::package(self, package: impl AsRef<str>) -> Self
-pub fn rustdoc_json::Builder::target(self, target: String) -> Self
-pub fn rustdoc_json::Builder::target_dir(self, target_dir: impl AsRef<Path>) -> Self
-pub fn rustdoc_json::Builder::toolchain(self, toolchain: impl Into<Option<String>>) -> Self
-pub fn rustdoc_json::build(options: rustdoc_json::Builder) -> Result<PathBuf, rustdoc_json::BuildError>
+pub fn rustdoc_json::Builder::features<I: core::iter::traits::collect::IntoIterator<Item = S>, S: core::convert::AsRef<str>>(self, features: I) -> Self
```
Emilgardis, line 33: this would ideally be rendered `core::iter::IntoIterator` as mentioned here https://github.com/Enselic/cargo-public-api/issues/183#issuecomment-1279896449

Emilgardis (approved):

rustdoc-json/public-api.txt
Enselic: We could have an internal translation map to do such adjustments

But currently there it looks like we can't merge this PR: https://github.com/Enselic/cargo-public-api/issues/183#issuecomment-1279933727

Thanks for review though

Enselic:
Looking at the discussion in https://github.com/rust-lang/rust/pull/103085, the probability seems high that this the `paths` that we rely on will not be removed until a proper fix (all items are fully present in the JSON) is in place.

I also just made a release, so we'll have time to let this sit in git `main` for a while and get some casual usage before hitting a release.

So I feel comfortable to merge this now.

PR #186, opened 2022-10-16
Fix `version.workspace = true` crash by bumping `cargo_toml` to `0.13.0`

### LINKED ISSUE #175: Toml parsing fails if it uses workspace keys

Enselic:
Closes #175

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -117,12 +117,11 @@ dependencies = [
 
 [[package]]
 name = "cargo_toml"
-version = "0.11.5"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


SUPERCILEX (approved):
🎉

Emilgardis (approved):
lgtm!

Issue #37, opened 2022-04-08
Updated sort order


douweschulte:
To keep related items instead of the current related kinds together it is proposed to sort based on the item path instead of the item string representation. See https://github.com/Enselic/public_items/pull/53#discussion_r846355796 for a possible implementation and more discussion.

Current situation:
```
- pub enum path::one::enum
- pub fn path::extra::enum::do_something()
- (possibly more fn)
- pub fn path::one::enum::do_something()
```

Proposed sorting
```
- pub fn path::extra::enum::do_something()
- pub enum path::one::enum
- pub fn path::one::enum::do_something()
- (possibly more fn)
```

Enselic:
To give some historical context: I have tried to base sort order on path before, but ended up thinking that the output became harder to read due to items not being neatly left-aligned in the same way as before. I think we can solve that better with a tree-based structure though. Maybe even indenting some parts of the output.

douweschulte:
That is a good thing to keep in consideration, I overlooked that. I do agree that the tree based output is the way forward and the sort order should be discussed at that moment.

Enselic:
In https://github.com/Enselic/cargo-public-api/pull/195 there is now proper grouping of all items, if the `--unsorted` flag is used. See example in that PR.

Here is my plan (subject to change):
1. Land that PR
2. Make a follow-up PR that by default uses the grouping rather than lexically sorting. Or put differently: The grouping will be considered being sorted already. That PR will close this issue.

Enselic:
PR that I *think* fixes this: https://github.com/Enselic/cargo-public-api/pull/198

PR #192, opened 2022-10-19
List `impl`s directly after corresponding items

Enselic:
We still only support auto-trait `impl`s, but more `impl`s will be supported later.

This draft illustrates what #189, #190 and #191 enables us to do. Still needs some polish and fixing, but I need to stop for today.

This PR takes us closer to solving #36, #37 and #39.


Enselic:
Obsoleted by https://github.com/Enselic/cargo-public-api/pull/195

PR #195, opened 2022-10-30
Render all `impl` items themselves

### LINKED ISSUE #36: Show Impls instead of all functions

### LINKED ISSUE #39: Present associated items as part of `impl`s

Enselic:
This is a rewrite on how items are processed.

Closes  #36 
Closes #39 

#37 will be closed after a follow-up PR that by default uses the logical grouping of items rather than lexical sorting.

(But I am very open to hearing opposing views on this)

<details>
  <summary>Example output (Click to expand)</summary>
  
```
$ cargo run -- --manifest-path test-apis/example_api-v0.1.0/Cargo.toml --unsorted --with-blanket-implementations
pub mod example_api
pub struct example_api::Struct
pub struct field example_api::Struct::v1_field: usize
impl RefUnwindSafe for example_api::Struct
impl UnwindSafe for example_api::Struct
impl Unpin for example_api::Struct
impl Sync for example_api::Struct
impl Send for example_api::Struct
impl<T, U> TryFrom<U> for example_api::Struct where U: Into<T>
pub type example_api::Struct::Error = Infallible
pub fn example_api::Struct::try_from(value: U) -> Result<T, <T as TryFrom<U>>::Error>
impl<T> Any for example_api::Struct where T: 'static + Sized
pub fn example_api::Struct::type_id(&self) -> TypeId
impl<T> From<T> for example_api::Struct
pub fn example_api::Struct::from(t: T) -> T
impl<T> BorrowMut<T> for example_api::Struct where T: Sized
pub fn example_api::Struct::borrow_mut(&mut self) -> &mut T
impl<T, U> Into<U> for example_api::Struct where U: From<T>
pub fn example_api::Struct::into(self) -> U
impl<T> Borrow<T> for example_api::Struct where T: Sized
pub fn example_api::Struct::borrow(&self) -> &T
impl<T, U> TryInto<U> for example_api::Struct where U: TryFrom<T>
pub type example_api::Struct::Error = <U as TryFrom<T>>::Error
pub fn example_api::Struct::try_into(self) -> Result<U, <U as TryFrom<T>>::Error>
impl Debug for example_api::Struct
pub fn example_api::Struct::fmt(&self, f: &mut $crate::fmt::Formatter<'_>) -> $crate::fmt::Result
pub fn example_api::function(v1_param: example_api::Struct)
```

</details>

Any input greatly appreciated, as always :)


douweschulte:
Looks very nice, the new `unsorted` option also looks really neat.

Enselic:
Thanks a lot for taking a look.

I plan on removing `--unsorted` in an upcoming PR (and instead make the output "good" by default), so I will not add tests specifically for `--unsorted` at this point.

Merging.

PR #196, opened 2022-10-31
Support `--diff-published your-crate@1.2.3` to diff against crates.io

### LINKED ISSUE #124: Be able to diff workspace crate against published crate version
It would be great to have an alternative to `--diff-git-checkouts` that instead of syncing a previous commit of the same repo (which can be a quite heavy and complex operation on a large repo) you can compare against a previously published crate version on crates.io. Have there been any thoughts of such an alternative?

So something like `--diff-published-crate x.y.z` which downloads the crate from crates.io (or the registry it was specified to use), unpacks it, and compares it from that.

We have a monorepo that has `rust-toolchain.toml` and lots of forked dependencies so syncing previous git versions and comparing that can be a bit problematic and slow. Also makes it harder if you have pending staged changes etc.

Enselic:
Closes #124 

Example usage in our own git (note how it handles a crate part of a workspace) (output shortened for brevity):

```
$ cargo run -- -p rustdoc-json --diff-published rustdoc-json@0.4.2 --verbose         
Processing "/Users/martin/.cargo/cargo-public-api_build-root-for-published-crates/rustdoc-json-0.4.2/target/doc/rustdoc_json.json"
Processing "/Users/martin/src/cargo-public-api/target/doc/rustdoc_json.json"
Removed items from the public API
=================================
-pub const fn rustdoc_json::BuildOptions::all_features(self, all_features: bool) -> Self
-pub fn rustdoc_json::BuildOptions::target_dir(self, target_dir: impl AsRef<Path>) -> Self
-pub fn rustdoc_json::BuildOptions::toolchain(self, toolchain: impl Into<Option<String>>) -> Self

Changed items in the public API
===============================
(none)

Added items to the public API
=============================
+impl Unpin for rustdoc_json::BuildOptions
+impl UnwindSafe for rustdoc_json::BuildOptions
```

### DIFF
--- a/cargo-public-api/src/published_crate.rs
+++ b/cargo-public-api/src/published_crate.rs
@@ -1,0 +1,110 @@
+//! Creates a dummy project with a dependency on the crate we want to build
+//! rustdoc JSON for. We then build rustdoc JSON for the crate using this dummy
+//! project.
+
+use crate::Args;
+use anyhow::{anyhow, Result};
+use std::{fmt::Display, path::PathBuf};
+
+pub fn build_rustdoc_json(package_spec_str: &str, args: &Args) -> Result<PathBuf> {
+    let spec = PackageSpec::try_from(package_spec_str)?;
+
+    let build_dir = build_dir(&spec);
+    std::fs::create_dir_all(&build_dir)?;
+
+    let write_file = |name: &str, contents: &str| -> std::io::Result<PathBuf> {
+        let mut path = build_dir.clone();
+        path.push(name);
+        std::fs::write(&path, contents)?;
+        Ok(path)
+    };
+
+    write_file("lib.rs", "// empty lib")?;
+    let manifest = write_file("Cargo.toml", &manifest_for(&spec))?;
+
+    let builder = crate::builder_from_args(args)
+        .manifest_path(&manifest)
+        .package(&spec.name);
+    crate::build_rustdoc_json(builder)
+}
+
+/// Prefer a non-temporary dir so repeated builds can be incremental.
+fn build_dir(spec: &PackageSpec) -> PathBuf {
+    let mut build_dir = dirs::cache_dir().unwrap_or_else(std::env::temp_dir);
+    build_dir.push("cargo-public-api");
+    build_dir.push("build-root-for-published-crates");
+    build_dir.push(spec.as_dir_name());
+    build_dir
+}
+
+fn manifest_for(spec: &PackageSpec) -> String {
+    format!(
+        "\
+        [package]\n\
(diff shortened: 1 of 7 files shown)


cargo-public-api/src/published_crate.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+}
+
+pub fn build_rustdoc_json(package_spec_str: &str, args: &Args) -> Result<PathBuf> {
+    let spec = PackageSpec::try_from(package_spec_str)?;
+
+    // We don't need to preserve tempdir because we just want the cached crate in $HOME
+    let mut temp_package = build_dir_root();
+    temp_package.push("cargo-public-api_build-root-for-published-crates");
```
Emilgardis:
```suggestion
    temp_package.push("cargo-public-api");
```

Emilgardis (commented):
I think using a shared target dir makes sense here

cargo-public-api/src/published_crate.rs
Enselic: Thanks for taking a look at the code; always appreciated. May I ask why you want to change to just `cargo-public-api `? My reasoning for using `cargo-public-api_build-root-for-published-crates` is that it makes it much easier for users to find the project AND the code (via git grep) that is responsible for putting stuff in their `~/.cargo` dir. The risk is high they will discover that dir out-of-context, i.e. long after having used this feature.

Enselic:
> I think using a shared target dir makes sense here

Thanks for the input. I wonder how well that will work in practice though? Since we use the nightly compiler, I don't think there is any guarantee about forward compatibility? I.e. that a new version of nightly will properly handle partial build artifacts from an old compiler/rustdoc? And what if the user specifies incompatible crate versions? Either directly or indirectly. Will the compiler/rustdoc always sort that out? Including in the face of build.rs scripts that builds e.g. C++ libraries?

I think it will be tricky for users to diagnose incremental build problems, since the building happens under the hood and in a "mysterious" way. So I think it might make sense to be a bit defensive here.

cargo-public-api/src/published_crate.rs
Emilgardis: my thinking was similar to how `cargo-audit` works, but maybe

```suggestion
    temp_package.join("cargo-public-api/build-root-for-published-crates");
```

then, I agree with the reasoning, just not the long dir name :)

Enselic: Great idea! Let's use that 👍

Enselic:
I ended up using the user cache dir instead of the .cargo dir, but I used your proposed dir structure. I think using the cache is better though.

Merging!

PR #198, opened 2022-11-06
Group items logically when sorting them

### LINKED ISSUE #37: Updated sort order

Enselic:
Closes #37 

I also added a (hidden) `--with-indentation` option to make the output easier to read.

To see how items are grouped after this PR, I recommend this command:

```
cargo run -- --manifest-path test-apis/comprehensive_api/Cargo.toml --with-indentation
```

I need to go through the bless output in more detail when I have time, but I *think* this is a complete fix.

(I split up blessing and the fix in two separate commits to make it easier to review this PR.)


public-api/src/intermediate_public_item.rs
```diff
@@ -66,3 +80,53 @@ impl<'c> IntermediatePublicItem<'c> {
         context.token_stream(self)
     }
 }
+
+/// In order for items in the output to be nicely grouped, we add a prefix to
+/// each item in the path to an item. That way, sorting on the name (with this
+/// prefix) will group items. But we don't want this prefix to be be visible to
+/// users of course, so we do this "behind the scenes".
+pub(crate) fn sorting_prefix(item: &Item) -> &'static str {
```
douweschulte: Couldn't you return the integer directly instead of creating a string? I would think the sort order would be retained in that way and I think that way is a bit neater.

douweschulte:
Very nice work I would agree that it closes #37.

public-api/src/intermediate_public_item.rs
Enselic: Thanks for the suggestion, sounds good 👍 

Implemented as part of the just-pushed set of tweaks and fixes

Enselic: Just a note: It was not actually a string that was returned, only a &str, but I still think it is a good idea to use numbers, because when debugging with --debug-sorting, the paths line up much better, and it is slightly easier to maintain

PR #199, opened 2022-11-07
Fix crash with `package.edition.workspace = true`

Jake-Shadle:
This fixes an issue where, if a package used workspace inheritance in any of the `package` fields, the call to package_name would fail, as cargo_toml was used to deserialize the manifest, but if workspace keys are declared, it will fail if it hasn't been given the virtual manifest that the keys are inherited from. The easiest fix was to just...not use cargo_toml for this, as only the `name` field was being read, so we just manually (well via serde) deserialize just the `package.name`


Enselic:
Thanks for reporting and thanks for the proposed fix! Looks reasonable on first glance. Before I do a detailed review, can you also add a regression test for the fix please? There are tests for workspaces already (see https://github.com/Enselic/cargo-public-api/tree/main/cargo-public-api/tests/virtual-manifest), so it should be quite straightforward to add a test for this, I think.

Jake-Shadle:
Ok, so I had initially thought it was for any `package.<>.workspace = true` fields, but the existing test case was already setting `version.workspace = true`, when I looked a bit closer it turns out that _currently_ the bug will only show up if `package.edition.workspace = true` is used, because of this block https://gitlab.com/crates.rs/cargo_toml/-/blob/3c9c20e07638dba7f7ba462f6d5f9247146cd93b/src/cargo_toml.rs#L366-388, which is the reason this wasn't previously causing test failures.

Enselic:
Thanks for the update! Looking at your code, I figured we might as well remove `cargo_toml` entirely. I also wanted to re-arrange virtual manifests a bit and put them under ./test-apis/.

I created a PR into your branch: https://github.com/EmbarkStudios/cargo-public-api/pull/1

Do you think it looks OK?

Enselic:
(PR title adjusted since it will be used in auto-generated release-notes)

Enselic:
FYI: I started using `cargo-manifest` instead of our own code in https://github.com/Enselic/cargo-public-api/pull/208

Since all regression tests pass, it should be fine. But if something breaks, I wanted to give you a hint on what could be the cause.

PR #202, opened 2022-11-14
Update clap to v4

dnaka91:
Noticed that the project is still using clap v3, and tried updating it to v4. I hope I captured all the special configs right, some local tests all worked out fine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,32 +124,39 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "clap"
-version = "3.1.3"
+version = "4.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86f8c0e2a6b902acc18214e24a6935cdaf8a8e34231913d4404dcaee659f65a1"
+checksum = "0eb41c13df48950b20eb4cd0eefa618819469df1bffc49d11e8487c4ba0037e5"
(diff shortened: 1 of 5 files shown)


Enselic:
Looks great, thanks! Throughout the project I have been careful to ensure that there is a good set of regression tests in place. And since CI is green, I am confident nothing major broke 👍

dnaka91:
Thank you for the quick merge! Saw the many tests, which are really great. Should do that degree of regression tests on my own projects :sweat_smile:.

If anything turns out to be broken later, just ping me, and I'll have a look.

Enselic:
Thank you for the offer 🙏 

Do you happen to know how to make clap 4 wrap --help text paragraphs?

With clap 3 we had:

```
        --with-blanket-implementations
            Raise this flag to make items part of blanket implementations such as `impl<T> Any for
            T`, `impl<T> Borrow<T> for T`, and `impl<T, U> Into<U> for T where U: From<T>` be
            included in the list of public items of a crate.
```

With clap 4 we have:

```
      --with-blanket-implementations
          Raise this flag to make items part of blanket implementations such as `impl<T> Any for T`, `impl<T> Borrow<T> for T`, and `impl<T, U> Into<U> for T where U: From<T>` be included in the list of public items of a crate.
```

This is on Mac in Terminal.app when I do `cargo run -- --help`

dnaka91:
Ah yes, that is now hidden behind a feature flag. Maybe was even before, but is now an opt-in feature. I'll create another PR to enable it 🙇‍♂️

PR #203, opened 2022-11-14
Fix currently reported clippy lints

dnaka91:
I guess the title says it all, just fixed all warnings reported by the latest clippy.


Enselic:
Thanks!

PR #205, opened 2022-11-15
Enable clap help message text wrapping

dnaka91:
As discussed in #202, re-enabling the text wrapping feature for `clap` help messages. This was the default in clap v3, but is now an opt-in feature in clap v4.


Enselic:
Thanks again!

dnaka91:
Thank you for adding a regression test :heart:

PR #207, opened 2022-11-16
Print -h and --help as part of CI

Enselic:
So that it is easier to review the impact an PR has on the help texts.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,6 +31,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: RUSTDOCFLAGS='--deny warnings' cargo doc --locked --no-deps --document-private-items
+      - run: cargo run --locked -- -h
+      - run: cargo run --locked -- --help
 
   cargo-clippy:
     name: cargo clippy -- --deny clippy::all --deny clippy::pedantic ...
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -30,6 +30,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v2
       - run: RUSTDOCFLAGS='--deny warnings' cargo doc --locked --no-deps --document-private-items
       - run: cargo run --locked -- -h
       - run: cargo run --locked -- --help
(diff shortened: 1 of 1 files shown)


Emilgardis:
just a thought, I think it'd be nice to include the help output in the readme, and that way it can be tested for too.

```rs
use std::{
    fmt::Write,
    fs,
    io::{BufReader, Read},
    path::PathBuf,
};

use clap::Parser;
use pretty_assertions::assert_eq;

static MANIFEST_PATH: &str = env!("CARGO_MANIFEST_DIR");

fn assert_lines_contains<'a>(iter: impl Iterator<Item = &'a str>, items: &[&str]) -> bool {
    if items.is_empty() {
        return false;
    }
    let mut index = 0;

    for iter_item in iter {
        if iter_item == items[index] {
            index += 1;
            if index == items.len() {
                return true;
            }
        }
    }

    panic!("iterator does not contain {:?}", items[index]);
}

#[test]
fn readme_help_usage() {
    let opts = crate::Opts::try_parse_from(&["cargo-public-api", "--help"]);
    let mut usage_help = String::new();
    write!(usage_help, "```text\n{}\n```\n", opts.unwrap_err()).unwrap();
    usage_help = usage_help
        .lines()
        .map(|s| s.trim_end().to_owned() + "\n")
        .collect();
    println!("{:?}", usage_help);
    let mut readme = String::new();
    BufReader::new(
        fs::File::open(PathBuf::from(MANIFEST_PATH).join("README.md"))
            .expect("could not open README.md"),
    )
    .read_to_string(&mut readme)
    .expect("can't read README.md");

    assert_lines_contains(
        readme.clone().lines(),
        &[
            "<!--BEGIN commandline options-->",
            "<!--END commandline options-->",
        ],
    );

    let readme_usage_help: String = readme
        .split_inclusive('\n')
        .skip_while(|line| !line.starts_with("<!--BEGIN commandline options-->"))
        .skip(1)
        .take_while(|line| !line.starts_with("<!--END commandline options-->"))
        .collect();
    assert_eq!(usage_help, readme_usage_help);
}
```

is how I've done it before

Enselic:
Thank you for the proposal and for the example code. That seems like a nice way of doing it.

Not entirely sure I want to put the help texts in the README though.

Here is how I reason about it:

The target audience for the README is "busy people that have a lot to do and very little time and that want to quickly learn what cargo-public-api can do". For this audience, it is good to have a README that is small. I think I want to try to make it even smaller at some point.

The target audience for the help text is "people that have decided that want to use the tool, and now want to get a complete list of its capabilities". So the target audience is quiet different, in my view.

This is just how I think about it. I could of course be wrong and confused and would be happy to discuss this further if you think I am missing something.

Enselic:
Inspired by your idea, https://github.com/Enselic/cargo-public-api/pull/237 adds the help output to docs/, and also adds test for the help output.

PR #209, opened 2022-11-19
Allow omitting package name when diffing against crates.io

Enselic:
So instead of `--diff-published crate-name@1.2.3` you can now just do `--diff-published @1.2.3`. The `crate-name` part will be taken from `Cargo.toml` (or `--package ...` arg).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,6 +100,7 @@ dependencies = [
  "anyhow",
  "assert_cmd",
  "atty",
+ "cargo-manifest",
  "cargo_metadata",
  "clap",
  "diff",
--- a/cargo-public-api/Cargo.toml
+++ b/cargo-public-api/Cargo.toml
@@ -15,6 +15,7 @@ repository = "https://github.com/Enselic/cargo-public-api"
 nu-ansi-term = "0.46.0"
 anyhow = "1.0.53"
 atty = "0.2.14"
+cargo-manifest = "0.4.0"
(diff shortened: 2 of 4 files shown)


Emilgardis:
I think for this to make more sense the `@` should be optional too

Enselic:
Good point, I agree that makes sense to support for `--diff-published` 👍

But what do we do with `--diff` in that case? Does `cargo public-api --diff 1.2.3` diff against the git tag or against the published crate? In theory it will not matter, but I don't think we can rely on that.

I do think it might make sense to consider "published diffing" to be the default, because it is both very relevant and less intrusive than than doing git checkouts.

I am contemplating changing to `cargo public-api diff ...`. Maybe that would be a good opportunity to change the default interpretation.

Emilgardis:
if there's an ambiguity, then it could error that it can't decide what to do. you'd then use `@` for crates.io/crate registry version, unsure what symbol tags would use

Enselic:
> if there's an ambiguity, then it could error that it can't decide what to do. you'd then use @ for crates.io/crate registry version, unsure what symbol tags would use

Sounds good to me :+1: 
We could require users to use `--diff-git-checkouts` for tags in that case

In the spirit on doing many small incremental steps, would you be OK with us merging this, and then someone can at some point do the "without @" simplification as future follow-up PR?

Emilgardis (approved):

PR #210, opened 2022-11-20
Turn `--diff` into just `diff` (make it a subcommand)

Enselic:
This passes CI and is a rather complete solution, but I still would like to call it a prototype.

This changes from

    cargo public-api --diff...

to

    cargo public-api diff...

Examples: Old vs new:
---

Diffing checkouts:

    cargo public-api --diff-git-checkouts v0.1.0 v0.2.0
    cargo public-api diff --git-checkouts v0.1.0 v0.2.0

Diffing against crates.io, short form:

    cargo public-api --diff @0.1.0
    cargo public-api diff @0.1.0

Pros of subcommand:
---

* Slightly easier to write

* clap enforces that `--deny` can only be used when diffing

Cons:
---

* Generic args can't be used after `diff`. This works: `cargo public-api --diff @0.1.0 --simplified`, but this does not work: `cargo public-api diff @0.1.0 --simplified`

* `cargo public-api --help` does not show `diff --help`. Instead `cargo public-api diff --help` must be invoked.

TODO:
---
* Would be good to keep old commands working, but printing a deprecation warning and referring to new syntax
* Tweak `--help` and `diff --help` output
* Decide if we want this. I think I am in favor of doing this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ which will print the public API of `regex` with one line per public item in the 
 
 ## Diff the Public API
 
-To diff the API between say **0.2.2** and **0.2.3** of `regex`, use `--diff-git-checkouts 0.2.2 0.2.3` while standing in the git repo. Like this:
+To diff the API between say **0.2.2** and **0.2.3** of `regex`, use `diff --git-checkouts 0.2.2 0.2.3` while standing in the git repo. Like this:
 
 ```bash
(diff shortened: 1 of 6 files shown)


Enselic:
PR obsoleted by https://github.com/Enselic/cargo-public-api/pull/235

PR #211, opened 2022-11-21
Run feature tests in parallel

Enselic:
For me this makes repeated runs of `time cargo test -p cargo-public-api` finish in 3.2s instead of 4.5s.

(Ignore whitespace changes to get a nice diff.)


Emilgardis (approved):
nice!

PR #212, opened 2022-11-21
Bump minimum nightly version to `nightly-2022-11-22`

Enselic:
Adapt to  https://github.com/rust-lang/rust/pull/104499 being merged

This PR should fail with todays nightly but pass with tomorrows nightly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -716,9 +716,9 @@ dependencies = [
 
 [[package]]
 name = "rustdoc-types"
-version = "0.18.0"
+version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Enselic:
I will wait with merging this, because due to the nature of the change, the code in `main` works both with yesterdays nightly and today's nightly.

By waiting with merging, more people will have the right minimum nightly version installed already when we bump the minimum version.

Enselic:
Actually, why merge this at all?

Let's re-open this next time we need to bump nightly.

Enselic:
[Another rustdoc JSON format change](https://github.com/rust-lang/rust/pull/106354) has been queued to be merged, so it is time to merge this now.

PR #213, opened 2022-11-23
Build more tests with independent target dirs

Enselic:
Before this commit many tests share target dirs, which is bad for parallelism because of `./target/debug/.cargo-lock`.

On my 16-core machine this commit makes

    time cargo test -p cargo-public-api

take 2.7s instead of 3.3s, and fixes all (knock on wood) test flakiness.


Enselic:
Just wanted to make sure nothing broke on Windows and Mac. Merging.

PR #214, opened 2022-11-23
Enable all tests on Windows in CI

Enselic:
I wonder if tests magically work on Windows in CI now given the recent changes to tests?

### DIFF
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -45,9 +45,7 @@ fn create_test_repo_with_dirty_git_tree() -> TestRepo {
     test_repo
 }
 
-// FIXME: This tests is ignored in CI due to some unknown issue with windows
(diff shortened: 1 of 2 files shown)


Enselic:
Amazingly they did 🤩
Almost seems too good to be true.

@Emilgardis Am I missing something? 🤔 (Asking you since IIRC you tried to debug these failures in the past)

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -47,7 +47,6 @@ fn create_test_repo_with_dirty_git_tree() -> TestRepo {
 
 // FIXME: This tests is ignored in CI due to some unknown issue with windows
```
Enselic: Oops I should also remove the FIXMEs

Emilgardis:
well well, seems to work now?!

Enselic:
Yep :shrug:

PR #215, opened 2022-11-24
Replace `test-invocation-variants.sh` with `cargo test`

Enselic:
This makes `./scripts/run-ci-locally.sh` run in ~10s instead of ~12s for me.

It also significantly simplifies the testing infrastructure.


Enselic:
I think the problem we are seeing is https://github.com/rust-lang/rustup/issues/988

We might need to handle `rustup` concurrency  ourselves somehow :(
Or simply install toolchains as a pre-`cargo test` step (:vomiting_face: )

An adventure for some other day...

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -766,10 +840,24 @@ impl TestCmd {
...
     }
 
-    fn new_impl(as_subcommand: bool, simplified: bool) -> Self {
+    fn with_toolchain(mut self, toolchain: &str) -> Self {
+        ensure_toolchain_installed(toolchain);
+        self.cmd.arg("--toolchain").arg(toolchain);
+        self
+    }
+
+    fn new_impl(toolchain: Option<&str>, as_subcommand: bool, simplified: bool) -> Self {
```
Emilgardis: this could be an enum to remove the assert

```rust
enum TestCmdArg<'str> {
	Toolchain(&'a str),
    Bin,
    Subcommand,
}
```

or

```rust
enum TestCmdArg<'str> {
	Subcommand(Option<&'a str>),
    Bin,
}
```

```suggestion
    fn new_impl(arg: impl Into<TestCmdArg<'_>>, simplified: bool) -> Self {

Comment (Enselic) [reply cargo-public-api/tests/cargo-public-api-bin-tests.rs]: Good idea :+1:. Fixed (but I called it `TestCmdType`). (The use of `&` is because pedantic clippy insisted.)

I also figured out a workaround for the `rustup` problem, I think.

Looks like the `cargo` binary on `windows-latest` is not the proxy version of `cargo` though:

> Did you mean to invoke `cargo` through `rustup` instead?

so that also needs to be solved somehow . Maybe also related to the way we use `rustup` in the tests...

File: cargo-public-api/tests/cargo-public-api-bin-tests.rs [hunk 2]
```diff
@@ -766,14 +850,24 @@ impl TestCmd {
...
+        ensure_toolchain_installed(toolchain);
+        self.cmd.arg("--toolchain").arg(toolchain);
+        self
+    }
+
+    fn new_impl(cmd_type: &TestCmdType, simplified: bool) -> Self {
+        let mut cmd = match cmd_type {
+            TestCmdType::Subcommand { toolchain } => {
+                test_utils::add_target_debug_to_path();
+                let mut cmd = Command::from_std(std::process::Command::new("cargo"));
```
Emilgardis, line 863: might want to do the same thing as here: https://github.com/Enselic/cargo-public-api/blob/e5324e4dca2c03d42a36d2cfd635322c52e6bf34/cargo-public-api/src/toolchain.rs#L10-L12

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -766,14 +850,24 @@ impl TestCmd {
...
+                if let Some(toolchain) = toolchain {
+                    cmd.arg(format!("+{}", toolchain));
+                }
+                cmd.arg("public-api");
```
Emilgardis: might want to do the same thing as here: https://github.com/Enselic/cargo-public-api/blob/e5324e4dca2c03d42a36d2cfd635322c52e6bf34/cargo-public-api/src/toolchain.rs#L10-L12

Enselic: Hmm yes that might work, but then the test would not do what regular users do, which IMHO is quite important.

After thinking some more about how to approach this short term, I think requiring toolchains  to be pre-installed before tests run is reasonable, and provide a helper script to do so. I'd really like to get this PR merge sooner rather than later, because otherwise the risk of conflicts is high, and I see it as a significant improvement.

If I just can get this to work in Windows now...

Enselic:
Hmm interesting, we get the same error now. So probably we don't need `./scripts/install-toolchains-for-tests.sh`. We just need the regular code to work on Windows.

Maybe time for me to buy a Windows license so I can debug this...

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Emilgardis: My bad, I should have made myself clear.

The code works around https://github.com/rust-lang/rustup/issues/3036, which is this exact issue (however the error message seems to have changed a bit)

```
error: no such subcommand: `+beta`
    
    Cargo does not handle `+toolchain` directives.
    Did you mean to invoke `cargo` through `rustup` instead?
```

Enselic: Thanks a ton for clarifying. I had completely forgotten about that bug. You saved me a lot of debugging time.

I've pushed a fix now, and CI is green.

I can think of many follow-up improvements to this PR, but let's get this merged before we improve on it further.

One idea:
Why don't we put the helper to install toolchains in e.g. `rustdoc-json` crate, and then install the necessary toolchain automatically for users (if it is missing) during regular `cargo public-api` use?

Emilgardis: I would not expect `rustdoc-json` to have rustup stuff in it. I'm definitely myself against installing the toolchain automatically via a `rustup toolchain install` but I see the value 👍🏼

Enselic: Interesting viewpoint! Let's break it down to a specific example. Consider a user that has `stable` installed but not `nightly`. Then that user does `cargo install cargo-public-api` and then `cargo public-api`.

In this case, I bet that say 95% of users want us to do the work to install `nightly` for them. Otherwise the tool that they just downloaded is useless!

The 5% that do NOT want us to install `nightly` for them can just Ctrl-C and stop the process.

This is a classic interaction design rule actually. Don't let the 5% ruin the experience for the other 95% :)

I am of course happy to keep discussing this. There might very well be some aspect here I am not thinking about. Meanwhile, I have created a new `rustup-toolchain` crate: https://crates.io/crates/rustup-toolchain, which we'll need it regardless of if we install `nightly` by default or allow users to `opt-in`. I'll wait with creating a PR for that though until we have landed this PR.

Emilgardis: good point, I guess I'm that 5% :D

Enselic: I did a quick experiment, but all tests started failing after uninstalling `nightly` toolchain, so I don't see me pursuing auto-install in the near future

For the record, prototype code at https://github.com/Enselic/cargo-public-api/compare/main...auto-install-toolchain

I mean, probably  quite easy to make it work, but I don't consider this urgent or crucial, so I'll put it on hold for now

PR #220, opened 2022-11-28
auto bless nightly 2022 11 28 1669659993

Enselic:
- Auto-create PR with blessed output from latest nightly
- Bless output with nightly-2022-11-28


Enselic:
sorry for the spam, just trying to get auto-blessing to work..

PR #230, opened 2022-11-29
Bless `nightly-2022-11-29` output

EnselicCICD:
Automatically created by https://github.com/Enselic/cargo-public-api/actions/runs/3578104957


Enselic:
Looks like this is finally working. There is one mac runner missing, but I don't want to wait for it. All other jobs a green so I don't see how that will not become green too.

Merging.

PR #240, opened 2022-12-11
Hide `--toolchain` from `-h`/`--help`

Enselic:
We want users to use `cargo +toolchain public-api` instead of `cargo public-api --toolchain toolchain`.

And we want the help output to be as free of noise as possible, to make it easier to digest for our users.


Emilgardis (approved):
Looks good!

PR #242, opened 2022-12-12
Support `--document-private-items`

dimpolo:
See https://doc.rust-lang.org/rustdoc/command-line-arguments.html#--document-private-items-show-items-that-are-not-public

### DIFF
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -24,6 +24,7 @@ use public_api::MINIMUM_RUSTDOC_JSON_VERSION;
 use tempfile::tempdir;
 use test_utils::assert_or_bless::AssertOrBless;
 use test_utils::rustdoc_json_path_for_crate;
+use test_utils::rustdoc_json_path_for_crate_with_private_items;
 
 #[path = "../src/git_utils.rs"] // Say NO to copy-paste!
 mod git_utils;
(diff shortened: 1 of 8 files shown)


Enselic:
Thank you for taking the time to create a PR :pray: 

Looks good! Except your code didn't work ;)

```
error: Found argument '--document-private-items' which wasn't expected, or isn't valid in this context

  If you tried to supply '--document-private-items' as a value rather than a flag, use '-- --document-private-items'

Usage: cargo rustdoc --lib --target-dir <DIRECTORY> --quiet --manifest-path <PATH> [args]...
```

I fixed that and added a regression test.

May I ask what your use case is? Since your code didn't work, it makes me suspect you don't need it by yourself, but instead you just thought this option could be good to have? I agree with that. I think it is only a matter of time before someone needs it.

dimpolo:
> Looks good! Except your code didn't work ;)

Ups sorry, thanks for catching that!

> May I ask what your use case is? 

Sure! I'd like to experiment with the possibility of compile-time introspection by abusing rustdoc 😅
The use case would be improving my [rusteval](https://github.com/dimpolo/rusteval) crate.
I'd like to have a build script or proc macro generate a dynamic interface for a chosen struct and use that to drive a REPL.

Enselic:
Very interesting! Good luck on your adventures!

I just released the new function in [rustdoc-json 0.7.2](https://crates.io/crates/rustdoc-json/0.7.2)

PR #246, opened 2022-12-15
Add `rustdoc_json::Builder::package_target()` to build for bin, test, bench, etc

### LINKED ISSUE #245: rustdoc-json: add support for binaries
Currently rustdoc-json uncoditionallly passes `--lib`. 
This causes the following error on binary only crates:
```shell
error: no library targets found in package `...`
```

I'm not sure what a nice interface would look like here, so I though I'd open this issue to discuss :)

One idea I had was to change the `Builder::package` parameter from a `impl AsRef<str>` to a

```rust
enum Package<T: AsRef<str>> {
    Bin(T),
    Lib(T),
}
```

dimpolo:
I'll see if I can write some tests tomorrow.

One annoyance I've noticed is that this doesn't work:
```rust
//  cannot infer type of the type parameter `T` declared on the enum `PackageTarget`
Builder::default().package_target(PackageTarget::Lib) 
```
Since `Lib` is the default, i think this won't be a problem very often.

Fixes #245


rustdoc-json/src/builder.rs
```diff
@@ -298,6 +330,22 @@ impl Builder {
     }
 }
 
+/// The part of of the package to document
+#[derive(Default, Debug, Clone)]
+pub enum PackageTarget<T: AsRef<str> = String> {
```
Enselic: Let's make this `#[non_exhaustive]` so we can add more variants in the future without having to do an incompatible version bump

rustdoc-json/tests/package_targets.rs
```diff
@@ -0,0 +1,24 @@
+use rustdoc_json::{Builder, PackageTarget};
+
+#[test]
+fn test_alternative_package_targets() {
```
Enselic: Can you move this to `rustdoc-json/tests/rustdoc-json-lib-tests.rs` please?

(Tests can run more in parallel if they are in a single binary)

rustdoc-json/tests/package_targets.rs
```diff
@@ -0,0 +1,24 @@
...
+    let targets = [
+        PackageTarget::Bin("comprehensive_api"),
+        PackageTarget::Bin("main2"),
+        PackageTarget::Test("test"),
+        PackageTarget::Example("example"),
+        PackageTarget::Bench("bench"),
+    ];
+
+    for target in targets {
+        let path = Builder::default()
```
Enselic: This test is currently a bit slow to run

It would be nice to build rustdoc JSON in parallel and independently. You can see https://github.com/Enselic/cargo-public-api/pull/211 for how we did it for another test

However, I don't think doing this is a blocker for merging this PR. I'm fine with if we wait with running these tests in parallel.

test-apis/comprehensive_api/benches/bench.rs
```diff
@@ -0,0 +1 @@
+fn main() {}
```
Enselic: I feels a bit off to me to put binaries and benchmarks in the `comprehensive_api` crate.

May I propose that you move this into a new crate only for rustdoc-json? Maybe `rustdoc-json/tests/test-crate` for example

rustdoc-json/src/builder.rs
```diff
@@ -177,6 +195,7 @@ impl Default for Builder {
             all_features: false,
             features: vec![],
             package: None,
+            package_target: PackageTarget::Lib,
```
Enselic:
```suggestion
            package_target: PackageTarget::default(),
```

rustdoc-json/src/builder.rs
```diff
@@ -298,6 +330,22 @@ impl Builder {
...
 }
 
+/// The part of of the package to document
+#[derive(Default, Debug, Clone)]
+pub enum PackageTarget<T: AsRef<str> = String> {
+    /// Document the package as a library, i.e. pass `--lib`
+    #[default]
+    Lib,
+    /// Document the given binary, i.e. pass `--bin <name>`
+    Bin(T),
```
Enselic: > One annoyance I've noticed is that this doesn't work:

I wonder if we need this to be generic? Wouldn't it be a bit nicer if we simply stored a `String` instead of a `T`?

The builder method can still take an `AsRef<str>`. Edit: Or maybe it can't? :thinking: Let's try

Enselic (commented):
Thank you, looks good overall! I added some comments.

rustdoc-json/src/builder.rs
dimpolo: Done

rustdoc-json/src/builder.rs
dimpolo: Do you mean adding methods like these?
```rust
impl PackageTarget {
    fn bin<T: AsRef<str>>(target: T) -> Self {
        ...
    }
}
```

test-apis/comprehensive_api/benches/bench.rs
dimpolo: I tried this. Could you have a look at the CI Error? I'm afraid I don't understand it.
It's suggesting to
>  add the package to the `workspace.exclude` array

but I did this already 🙄

rustdoc-json/src/builder.rs
Emilgardis: the methods should take either `T: Into<String>` or `T: ToString` if the generic is removed

rustdoc-json/tests/package_targets.rs
Enselic: (I pushed a fix for this now, so now these tests run in 0.5 seconds for me instead of 2 seconds)

test-apis/comprehensive_api/benches/bench.rs
Enselic: Fixed in the commit I just pushed

rustdoc-json/src/builder.rs
Enselic: I pushed a commit where I keep things simple and just let the enum own the String. I am a big fan of simple. Let me know what you think please (I need to quit coding for the evening)

Emilgardis (approved):
lgtm!

test-apis/comprehensive_api/benches/bench.rs
dimpolo: Thanks!

rustdoc-json/src/builder.rs
dimpolo: lgtm!

rustdoc-json/tests/package_targets.rs
dimpolo: awesome! thanks

dimpolo:
Thanks for the review and adjustments!

Enselic:
Thank for the help with implementing!

(Again, feel free to ping when you would like a release of rustdoc-json)

PR #247, opened 2022-12-18
Correctly determine JSON path for `Builder::package("foo@1.0.0")`

dimpolo:
Sorry for all the PRs 😅


Enselic:
Trust me, I feel nothing but appreciation whenever you open a PR 🙂

I'm a bit unsure about the use case here? Why would someone pass @ in the package name?

dimpolo:
Glad to hear that :)

When your dependency graph contains the same crate multiple times at different versions you have to specify which one you want.

Enselic:
But since rustdoc-json disregards the version, I don't think it makes sense to have an API that allows version to be specified? Or am I missing something?

dimpolo:
My understanding is this:

`builder.package("cratename")` might fail if there's multiple `cratename`s in the dependecy graph

`builder.package("cratename@1.0.0")` will correctly pass this to rustdoc and produce a json file

that json file will be named "cratename.json"

Enselic:
Thank you for explaining, that makes sense.

We're gonna need a regression test for this fix. (Otherwise the making of each release will induce maintainer angst  caused by fear of breaking this fix.) Probably easiest is to create two dependencies for `rustdoc-json/tests/test_crate`,  like `rustdoc-json/tests/test_crate_dep_1_0_0` and `rustdoc-json/tests/test_crate_dep_2_0_0`.

However, I wonder if there is not an upstream bug at the bottom of this. If the crate version is not part of the rustdoc JSON file output file name, one of the files will overwrite the other. Perhaps enough of an edge case to not be a problem in practice,however.

I wonder how the rustdoc HTML output files are named in a case like this.

rustdoc-json/tests/test_crates/test_dep_1_0_0/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "test_dep"
+version = "1.0.0"
+edition = "2021"
+
+# Exclude this package from the top level workspace by declaring our own empty
+# workspace
+[workspace]
```
Emilgardis: missing a newline

rustdoc-json/tests/test_crates/test_dep_2_0_0/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "test_dep"
+version = "2.0.0"
+edition = "2021"
+
+# Exclude this package from the top level workspace by declaring our own empty
+# workspace
+[workspace]
```
Emilgardis: ditto

rustdoc-json/tests/rustdoc-json-lib-tests.rs
```diff
@@ -53,3 +53,31 @@ fn test_alternative_package_target(package_target: PackageTarget) {
...
+        .manifest_path("tests/test_crates/test_crate/Cargo.toml")
+        .quiet(true) // Make it less noisy to run tests
+        .target_dir(&target_dir);
+
+    // test_dep is present multiple times in the dependency graph.
+    // Check that just passing "test_dep" errors
+    match builder.clone().package("test_dep").build() {
+        Err(rustdoc_json::BuildError::General(_)) => {}
+        _ => panic!("Expected ambiguous specification error"),
+    }
```
dimpolo, line 72: Note that this will print the following in the build output:
```
error: There are multiple `test_dep` packages in your project, and the specification `test_dep` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  test_dep@1.0.0
  test_dep@2.0.0
```
Feel free to remove this :)

rustdoc-json/tests/test_crates/test_dep_2_0_0/Cargo.toml
dimpolo: fixed :)

dimpolo:
> We're gonna need a regression test for this fix. (Otherwise the making of each release will induce maintainer angst caused by fear of breaking this fix.) Probably easiest is to create two dependencies for rustdoc-json/tests/test_crate, like rustdoc-json/tests/test_crate_dep_1_0_0 and rustdoc-json/tests/test_crate_dep_2_0_0.

Sure thing! I added a test.

> However, I wonder if there is not an upstream bug at the bottom of this. If the crate version is not part of the rustdoc JSON file output file name, one of the files will overwrite the other. Perhaps enough of an edge case to not be a problem in practice,however.

> I wonder how the rustdoc HTML output files are named in a case like this.

That's a good point. Running `cargo rustdoc` overwrites one instance of `test_dep`. Running `cargo doc` does the same but produces a warning:
```
warning: output filename collision.
The lib target `test_dep` in package `test_dep v2.0.0 (cargo-public-api\rustdoc-json\tests\test_crates\test_dep_2_0_0)` has the same out
put filename as the lib target `test_dep` in package `test_dep v1.0.0 (cargo-public-api\rustdoc-json\tests\test_crates\test_dep_1_0_0)`.
Colliding filename is: cargo-public-api\rustdoc-json\tests\test_crates\test_crate\target\doc\test_dep\index.html
The targets should have unique names.
This is a known bug where multiple crates with the same name use
the same path; see <https://github.com/rust-lang/cargo/issues/6313>.
```

Clickable link: https://github.com/rust-lang/cargo/issues/6313

dimpolo:
Looking at it closer, the warning seems to be targeted at our weird test setup.
In my testing setup my `cargo tree` includes this bit:
```
├── embedded-hal v0.2.7
│   │   ├── nb v0.1.3
│   │   │   └── nb v1.0.0
```
No warnings are produced here

Edit:
`cargo doc -p nb@1.0.0` => no warnings
`cargo doc -p nb@0.1.3` => warning

Fascinating, but I guess off topic 😅

Emilgardis:
do you have a virtual manifest and are not specifying `workspace.resolver = "2"` perhaps?

dimpolo:
No virtual manifests or workspaces involved. I think it's just this bug https://github.com/rust-lang/rust/issues/56169

rustdoc-json/tests/rustdoc-json-lib-tests.rs
Enselic: Thanks for the heads-up :+1: 

If we change from

```rust
    if *quiet {
        command.arg("--quiet");
    }
```

to

```rust
    if *quiet {
        command.arg("--quiet");
        command.stderr(std::process::Stdio::null());
    }
```

then the error goes away but all tests still pass.

I don't want to supress stderr just because of `quiet` however, so I need to figure out what the API should look like for this.

... if we even need an API for this at all ... probably only useful for tests ...

Enselic (approved):
Thanks, looks good! Setting Approved, but I'd like to figure out how to suppress the error when tests are run before we merge this PR, because it is quite distracting :thinking:

Enselic:
I think a good solution will be to add a `slient(true)` method on `Builder`, which would suppress stdout and stderr.

It will be non-trivial to write tests for that though, so I don't want to block this PR on it. Merging.

Merging this PR does not violate the "always releasable main branch" rule, since the problem we have only relates to the running of our test suite.

rustdoc-json/tests/test_crates/test_dep_1_0_0/Cargo.toml
Enselic: https://github.com/Enselic/cargo-public-api/pull/249 automates this part of the review :)

PR #248, opened 2022-12-20
Add `rustdoc_json::Builder::silent()` to suppress stdout and stderr

Enselic:
And use it to suppress stderr from `test_specified_dependency_version()` (see https://github.com/Enselic/cargo-public-api/pull/247#pullrequestreview-1223510330)

CC @dimpolo (and Emilgardis but I think you see this anyway)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -601,9 +601,11 @@ dependencies = [
 name = "rustdoc-json"
 version = "0.7.3"
 dependencies = [
+ "assert_cmd",
  "cargo-manifest",
  "cargo_metadata",
  "expect-test",
+ "predicates",
  "public-api",
  "serde",
  "tempfile",
--- a/rustdoc-json/CHANGELOG.md
+++ b/rustdoc-json/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
-* Add `Builder::package_target` and `PackageTarget`
+* Add `Builder::package_target()` and `PackageTarget`
 * Correctly determine json path for `Builder::default().package("crate@1.0.0")`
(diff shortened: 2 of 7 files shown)


Emilgardis (approved):
lgtm, however, would it maybe make sense to allow aribtrarily setting stderr and stdout to whatever the user wants?


So instead of `Builder::silent()`, it'd be `Builder::stderr(Stdio::null()).stdout(Stdio::null())`

Might be worth even making it possible the get the `Command`, so something like 

```rust
Builder::build_with_command(|&mut cmd| {
        cmd.stdout(std::process::Stdio::null());
        cmd.stderr(std::process::Stdio::null());
    })

impl Builder {
    fn build_with_command(f: impl FnOnce(&mut Command)) {
        todo!()
    }
}
```

Enselic:
> allow aribtrarily setting stderr and stdout to whatever the user wants?

I like this idea, and began to implement it, but it is quite tricky to do. For one, then we can't derive `Clone` any more. That is a small problem, but we also need to restructure `cargo_rustdoc_command()` so it gets ownership of the `Stdio` instead of borrowing it. Another downside is that `Stdio` becomes part of the public API, which is an external type. Even though it has been around since Rust 1.0.0, it still puts some limitation on how free we are to evolve the API.

I also thought about exposing the `Command` directly after you proposed it, but IMHO that would expose a way too big API surface. It would become very tricky to analyze what constitutes backwards compatibility and not. For example, if a client starts using some `rustdoc` JSON flag, we can't start using it ourselves too without breaking backwards compatibility (because the client code would stop working; `rustdoc` would likely complain about the same option used twice).

So in short, big thanks for the review and the proposals. But for now, let's go with what we have in the PR :)

PR #249, opened 2022-12-20
Ensure all files in the git repo ends with a newline char

Enselic:
Also move readme tests to a new repo-tests crate for increased test parallelism.

I want to automate as much as possible to make things as smooth as possible for maintainers and contributors, and we recently manually checked for trailing newlines: https://github.com/Enselic/cargo-public-api/pull/247#discussion_r1051671366

So let's automate :)


Emilgardis (approved):
good idea!

PR #250, opened 2022-12-20
Get rid of `enum variant` and `struct field` prefixes

Enselic:
They make the output look very non-Rusty, and by now I think the surrounding items (due to https://github.com/Enselic/cargo-public-api/pull/198 being merged some time ago) and the name of the item itself, makes it sufficiently clear what kind of item it is.

@douweschulte @Emilgardis : This is a quite significant change, so I would like to check with you if you think this sounds and looks good

### DIFF
--- a/cargo-public-api/tests/expected-output/diff_published.txt
+++ b/cargo-public-api/tests/expected-output/diff_published.txt
@@ -9,7 +9,7 @@ Changed items in the public API
 
 Added items to the public API
 =============================
-+pub struct field example_api::Struct::v2_field: usize
++pub example_api::Struct::v2_field: usize
 +pub struct example_api::StructV2
-+pub struct field example_api::StructV2::field: usize
++pub example_api::StructV2::field: usize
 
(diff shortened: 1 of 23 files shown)


douweschulte:
I do agree that they look quite weird from a Rust perspective. But do I get it correclty if you want to present the following:
```rust
pub struct pdbtbx::SequencePosition
pub struct field pdbtbx::SequencePosition::end: isize
pub struct field pdbtbx::SequencePosition::end_insert: core::option::Option<alloc::string::String>
pub struct field pdbtbx::SequencePosition::start: isize
pub struct field pdbtbx::SequencePosition::start_insert: core::option::Option<alloc::string::String>
```
as:
```rust
pub struct pdbtbx::SequencePosition
pub pdbtbx::SequencePosition::end: isize
pub pdbtbx::SequencePosition::end_insert: core::option::Option<alloc::string::String>
pub pdbtbx::SequencePosition::start: isize
pub pdbtbx::SequencePosition::start_insert: core::option::Option<alloc::string::String>
```
If this is the case I fully agree otherwise I would like to see how you envisioned it. Especially now with the grouping these items are close enough together that it is immediately clear what they are.

Enselic:
Yep exactly like that :+1:
(which can also be seen by looking at the diff of this PR)

Thanks for taking the time to give your input :pray: 

(Waiting a day or two more in case Emil wants to chime in)

Emilgardis:
I have no comment, I like the verbosity, but I also enjoy the simplicity :D

Enselic:
Thanks anyway for input. Let's merge this and see how it works out. We can always revert it, or make it opt-in, or opt-out, if needed :)

Issue #251, opened 2022-12-20
Group items even if multiple inherent `impl`s are used


douweschulte:
```rust
pub struct pdbtbx::PDB
pub struct field pdbtbx::PDB::identifier: core::option::Option<alloc::string::String>
... (more fields)
impl pdbtbx::PDB
impl pdbtbx::PDB
impl pdbtbx::PDB
impl<'a> pdbtbx::PDB
```

This example is from the crate `pdbtbx` where I have used multiple impl blocks on a single struct to allow fro grouping of functions with some general comments about those groups. That is why it stands out so much. But I would assume these impl blocks (just the main block with all attached functions) to not be shown in the list here. An argument could be made to have the functions grouped below their respective impl blocks, especially when generic bounds are used. But this seems to not be the case here.

Enselic:
Thank you for reporting :pray: 

I can see that in rustdoc HTML, each impl item is presented individually, together with the doc of the impl, if any. So I think we should also show each impl item individually.

So what I think we want is for this code:

```rust
pub struct MultipleInherentImpls;

impl MultipleInherentImpls {
    pub fn impl_one() {}
}

impl MultipleInherentImpls {
    pub fn impl_two() {}
}
```

to be rendered like this:

```txt
pub struct comprehensive_api::impls::MultipleInherentImpls
impl comprehensive_api::impls::MultipleInherentImpls
pub fn comprehensive_api::impls::MultipleInherentImpls::impl_one()
impl comprehensive_api::impls::MultipleInherentImpls
pub fn comprehensive_api::impls::MultipleInherentImpls::impl_two()
```

I looked into solving this. The tricky part is what to use to group items. If we use the impl Id for grouping, it works for listing, but not for diffing, since the ID changes. So I think we might need to do some non-trivial changes to item grouping to solve this.

Enselic:
On a second thought, if we collapse all inherent `impl`s into one `impl` item in the output, we don't need to solve the grouping. If we do it like that, the code above would render as:

```txt
pub struct comprehensive_api::impls::MultipleInherentImpls
impl comprehensive_api::impls::MultipleInherentImpls
pub fn comprehensive_api::impls::MultipleInherentImpls::impl_one()
pub fn comprehensive_api::impls::MultipleInherentImpls::impl_two()
```

The public API is (as far as I know) not affected by if inherent `impl`s are split up or not. So maybe it wouldn't be so bad to do that.  :thinking:

[...]

PR #252, opened 2022-12-21
Group impl blocks together with their respective functions

Emilgardis:


Emilgardis:
part of #251, does not solve the deduplication though

public-api/tests/expected-output/comprehensive_api.txt
```diff
@@ -139,13 +139,15 @@ pub comprehensive_api::structs::ConstArg::items: [T; N]
...
 pub fn comprehensive_api::structs::Plain::s1(self)
 pub fn comprehensive_api::structs::Plain::s2(&self)
 pub fn comprehensive_api::structs::Plain::s3(&mut self)
+impl<'a> comprehensive_api::structs::Plain
+impl<'a> comprehensive_api::structs::Plain
 pub fn comprehensive_api::structs::Plain::s4(&'a self)
+pub fn comprehensive_api::structs::Plain::s5(&'a self)
 pub struct comprehensive_api::structs::PrivateField
 pub struct comprehensive_api::structs::TupleStructDouble(pub usize, pub bool)
 pub struct comprehensive_api::structs::TupleStructDoubleWithHidden(_, pub bool)
```
Emilgardis, line 153: Ideally, this would be just one 

```
impl<'a> comprehensive_api::structs::Plain
```

public-api/src/intermediate_public_item.rs
```diff
@@ -29,22 +29,23 @@ impl<'c> NameableItem<'c> {
...
-            perceived_name = &trait_path.name;
+        if let ItemEnum::Impl(impl_) = &self.item.inner {
+            if let Some(trait_path) = &impl_.trait_ {
+                // In order for items of impls to be grouped together with its impl, add
+                // the "name" of the impl to the sorting prefix.
+                perceived_name = (&trait_path.name).into();
+            } else {
+                perceived_name =
+                    crate::tokens::tokens_to_string(&context.render_impl(impl_, &[])).into();
+            }
```
Emilgardis: is this the correct render?

rustdoc-json/public-api.txt
```diff
@@ -80,6 +80,12 @@ impl<T, U> core::convert::TryInto<U> for rustdoc_json::PackageTarget where U: co
 pub type rustdoc_json::PackageTarget::Error = <U as core::convert::TryFrom<T>>::Error
 pub fn rustdoc_json::PackageTarget::try_into(self) -> core::result::Result<U, <U as core::convert::TryFrom<T>>::Error>
 pub struct rustdoc_json::Builder
+impl core::clone::Clone for rustdoc_json::Builder
+pub fn rustdoc_json::Builder::clone(&self) -> rustdoc_json::Builder
+impl core::fmt::Debug for rustdoc_json::Builder
+pub fn rustdoc_json::Builder::fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result
+impl core::default::Default for rustdoc_json::Builder
+pub fn rustdoc_json::Builder::default() -> Self
 impl rustdoc_json::Builder
```
Emilgardis: its unclear to me why these have moved up

Emilgardis:
a nice diff: [diff](https://github.com/Enselic/cargo-public-api/pull/252/files/a574fc992ba8e94bf301c314e4bc0ae2571309de..4f1f714a319ff959e8540365399fd0e33ed90260#diff-f3ed37e2fac62e106f72b17db185bea4cd8155c0925f32e9699c04df1855a8b6)

rustdoc-json/public-api.txt
Emilgardis: solved!

Emilgardis:
It seems like deduplication is easily solved with `public_api.items.dedup_by_key(|i| i.to_string());` when sorted.

Enselic (approved):
Nice!

Feels a bit hacky to render the impl like this for sorting purposes. But much more importantly: the code works and there are regression tests. So with low risk and effort we will be able to implement it in some other way if we can think of a better way later.

Right now I can't think of a better way, so I definitely think we should merge this :+1:

Enselic:
> It seems like deduplication is easily solved with public_api.items.dedup_by_key(|i| i.to_string()); when sorted.

There is a relatively high risk with doing that: in case items due to bugs or limitations and up being rendered the same, using that method will cause items to disappear from listing/diffing. I quite strongly prefer items to occasionally be duplicated, over not being able to fully trust the output.

So I think we should do the deduplication  specifically for `impl`s somehow.

But maybe I am just overly pessimistic. Early on in the project I used the equivalent of what you propose, but that ended up having the problems I described. But nowadays the project is much more mature, so the risk of inappropriately identically rendered items is much lower.

Maybe we should be brave and go with your proposal. Because it is very simple. And I do like simple :)

(But let's iterate on this in a separate PR after this one is merged, as I think you also have planned already.)

Enselic:
The PR title sounds like a bugfix, so I'm changing label to category-bugfix. I hope that's OK.

I will make a release shortly.

PR #253, opened 2022-12-23
dedup items when sorted

Emilgardis:


Emilgardis:
See #251 and https://github.com/Enselic/cargo-public-api/pull/252#issuecomment-1363738209

public-api/src/lib.rs
```diff
@@ -222,6 +222,7 @@ impl PublicApi {
 
         if options.sorted {
             public_api.items.sort();
+            public_api.items.dedup_by_key(|i| i.to_string());
```
Enselic, line 225: Hmm right, this only works with sorted output. Spontaneously I think the set of items must not be different in sorted vs unsorted output.

I wonder how many users/clients that uses unsorted output though. One option could be to remove that option and always sort. Then there would be no inconsistencies at least.

Emilgardis:
superseded by https://github.com/Enselic/cargo-public-api/pull/270

PR #254, opened 2022-12-26
Upgrade all deps

SUPERCILEX:
I only really need cargo-manifest upgraded for this commit: https://github.com/LukeMathWalker/cargo-manifest/commit/bc3502ac9856aea49bd1faf92b27ab6fd0d647bd. Feel free to close if you'd rather do the upgrade.

### DIFF
--- a/public-api/src/item_processor.rs
+++ b/public-api/src/item_processor.rs
@@ -285,20 +285,21 @@ pub(crate) fn sorting_prefix(item: &Item) -> u8 {
         ItemEnum::AssocConst { .. } => 16,
 
         ItemEnum::Function(_) => 17,
+        ItemEnum::Method(_) => 18,
 
-        ItemEnum::Typedef(_) => 18,
+        ItemEnum::Typedef(_) => 19,
 
         ItemEnum::Impl(impl_) => match ImplKind::from(impl_) {
-            ImplKind::Normal => 19,
-            ImplKind::AutoTrait => 20,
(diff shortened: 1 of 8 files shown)


public-api/src/item_processor.rs
```diff
@@ -285,21 +285,20 @@ pub(crate) fn sorting_prefix(item: &Item) -> u8 {
         ItemEnum::AssocConst { .. } => 16,
 
         ItemEnum::Function(_) => 17,
-        ItemEnum::Method(_) => 18,
```
Enselic: To remain compatible with current minimun nightly version, I think we need to undo the changes in this file? Or do you need this change? If so, it is totally fine with me to bump minimum nightly version. Just let me know. I just want to avoid bumping unless it is necessary.

I prepared this change in https://github.com/Enselic/cargo-public-api/pull/212, but then realized it is not needed, because the current code is compatible with rustdoc JSON from both new and old nightlies.

Enselic (commented):
Thank you!

I pushed a commit to get rid of more deps, and also added a question.

Apart from that this looks good to me 🙏

public-api/src/item_processor.rs
SUPERCILEX: > Or do you need this change?

Nope, just wasn't aware of the background. :) Reverted.

Enselic: Great, let's merge this now then 👍

PR #260, opened 2022-12-28
Support `cargo public-api diff latest` to diff against the latest published version

Enselic:
Which will diff against the latest published version of a crate.

Downsides:
* Using `crates-index` adds 22 deps, crossing the 100 deps line (we end up at 115). Maybe we can find a more light-weight solution.

### DIFF
--- a/cargo-public-api/Cargo.toml
+++ b/cargo-public-api/Cargo.toml
@@ -25,6 +25,10 @@ thiserror = "1.0.38"
 version = "4.0.32"
 features = ["derive", "wrap_help"]
 
+[dependencies.crates-index]
+version = "0.18.11"
+default-features = false
+
 [dependencies.rustdoc-json]
 path = "../rustdoc-json"
 version = "0.7.4"
--- a/cargo-public-api/Cargo.toml
+++ b/cargo-public-api/Cargo.toml
@@ -11,6 +11,12 @@ categories = ["command-line-utilities", "development-tools::cargo-plugins"]
 license = "MIT"
 repository = "https://github.com/Enselic/cargo-public-api"
 
+[features]
+default = ["diff-latest"]
+# Enable `cargo public-api diff latest`. Requires 33 more deps (most notably the
+# `git2` crate) so that the crates.io index can be updated and read.
+diff-latest = ["crates-index"]
+
 [dependencies]
 nu-ansi-term = "0.46.0"
 anyhow = "1.0.68"
@@ -28,6 +34,7 @@ features = ["derive", "wrap_help"]
 [dependencies.crates-index]
 version = "0.18.11"
 default-features = false
+optional = true
 
 [dependencies.rustdoc-json]
 path = "../rustdoc-json"
--- a/cargo-public-api/Cargo.toml
+++ b/cargo-public-api/Cargo.toml
@@ -15,7 +15,7 @@ repository = "https://github.com/Enselic/cargo-public-api"
 default = ["diff-latest"]
 # Enable `cargo public-api diff latest`. Requires 33 more deps (most notably the
 # `git2` crate) so that the crates.io index can be updated and read.
-diff-latest = ["crates-index"]
+diff-latest = ["dep:crates-index"]
 
 [dependencies]
 nu-ansi-term = "0.46.0"
--- a/cargo-public-api/src/published_crate.rs
+++ b/cargo-public-api/src/published_crate.rs
@@ -2,7 +2,7 @@
 //! rustdoc JSON for. We then build rustdoc JSON for the crate using this dummy
 //! project.
 
-use crate::Args;
+use crate::{Args, LATEST_VERSION_ARG};
 use anyhow::{anyhow, Result};
 use std::path::PathBuf;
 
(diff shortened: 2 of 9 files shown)


Emilgardis:
```
cargo search --limit 1 <package>
```

is a neat way to get latest(where latest = highest) version

Enselic:
I tried it out and it seems to work, thanks!

The downside of that is that it puts load in the crates.io backend each time the diff subcommand is invoked (search results are probably not cached), which goes against crates.io [recommendations](https://crates.io/data-access). It will for sure not be a problem as long as `cargo public-api diff latest` has a small number of users. But if `cargo public-api diff latest` starts being used a lot, it could cause problems.

So I *think* I want to pursue the use of the `crates-index` crate a bit more. Maybe I can reduce its dependencies, for example.

But I would say `cargo search` still is an option, if `crates-index` doesn't work out in a satisfactory way. On the other hand, maybe not many users care about 22 new deps after all :thinking: 

Setting this as PR as draft until I have had time to explore this a bit more.

Emilgardis:
I don't think it's against the recommendations, it's not a crawl, it's the same thing as crates-index, but built in and with a worse api surface.

Enselic:
You are probably right that we would not be considered crawlers.

After having thought about this some more, I still think using `crates-index` is preferable however. Here is a comparison, roughly ordered by (IMHO) importance/weight in descending order:

| Aspect                                 | `cargo search`                                                                    | `crates-index`   | Winner         |
| -------------------------------------- | --------------                                                                    | --------------   | -------------- |
| Complexity in our code                 | Medium-complexity custom parsing code and tests                                   | Simple           | `crates-index` |
| `cargo-public-api` compile time        | No impact                                                                         | ~33% increase    | `cargo search` |
| Distinction between "latest" and "highest"  version| Not supported | Supported |  `crates-index` |
| Works offline (with caches primed)     | No                                                                                | Yes              |  `crates-index` |
| Architecture                           | Does not make use of the pre-distributed crates.io database every dev already has | Makes use of it. | `crates-index` |
| Number of added deps                   | 0                                                                                 | ~22~ 33               | `cargo search` |

We could turn this into an optional feature so that if a dev wants to iterate quickly on other parts of `cargo-public-api`, the dev can skip building ~22~ 33 deps. Unsure if it is worth the hassle though. What do you think? Had it been for the `public-api` library, then I would say for sure it should be an optional feature.

I haven't made up my mind 100%, so let's leave this PR as Draft for a while longer.

Emilgardis:
I think it's fine to use `crates-index` 👍🏼, also, I don't think it should be a concern

> We could turn this into an optional feature so that if a dev wants to iterate quickly on other parts of cargo-public-api, the dev can skip building 22 deps.

The only time these 22 (I count 34 however) new deps would be included in build time is on a fresh compilation.

Enselic:
Good catch, it is 33 deps. My methodology was wrong (I counted number of added deps in Cargo.lock)

It is true that in many cases the new deps will not need to be rebuilt. But for some workflows (in particular `git clean && git bisect`) it can be useful. While I do prefer simplicity and thus no need for features at all, 33 new deps seems to me to justify making a feature optional. Especially since it is a bit of an edge case (albeit in my eyes very useful) feature.

Since it was easy to do I ended up making this feature optional (but default). If it ends up causing a lot of trouble to e.g. maintain, we can make it non-optional later.

cargo-public-api/Cargo.toml
```diff
@@ -11,6 +11,12 @@ categories = ["command-line-utilities", "development-tools::cargo-plugins"]
 license = "MIT"
 repository = "https://github.com/Enselic/cargo-public-api"
 
+[features]
+default = ["diff-latest"]
+# Enable `cargo public-api diff latest`. Requires 33 more deps (most notably the
+# `git2` crate) so that the crates.io index can be updated and read.
+diff-latest = ["crates-index"]
```
Emilgardis:
```suggestion
diff-latest = ["dep:crates-index"]
```

this avoids having a feature `diff-latest` and `crates-index`

Enselic: Thanks, I didn't know about this actually :blush:

Emilgardis: `dep:` was introduced in 1.60, see https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies for more (where more is being able to do `my-feature = ["package-name?/feature-name"]` to only enable `package-name/feature-name` if `package-name` is also enabled)

Enselic:
(My intention is to (squash) merge this soon-ish; marking as Ready)

cargo-public-api/src/published_crate.rs
```diff
@@ -37,6 +44,26 @@ pub fn build_rustdoc_json(version: impl Into<String>, args: &Args) -> Result<Pat
...
+        })?;
+        let version = crate_.most_recent_version();
+        Ok(version.version().to_string())
+    }
+    #[cfg(not(feature = "diff-latest"))]
+    {
+        Err(anyhow!(
+            "Can't find latest version of `{package_name}`; the `diff-latest` feature is disabled"
+        ))
+    }
```
Emilgardis:
```suggestion
        Err(anyhow!(
            "Can not find latest version of `{package_name}`; the `diff-latest` feature needs to be enabled for `cargo-public-api`"
        ))
    }
```

Emilgardis (approved):
lgtm! 

Could probably also add it to the help text so that it's more discoverable


conditional documentation can be done with 

```rust
/// ...
#[cfg_attr(feature = "diff-latest", doc = "Can also take `latest` to diff current state with the last published version")]
```

Enselic:
> Could probably also add it to the help text so that it's more discoverable

Indeed :+1: I added it to the help text now. I don't think we need to only include it in case the feature is enabled though. Users get a helpful error message telling them how to make that feature work for them after all, and I do think they should know about this feature, even if they for some reason comes across a build with the feature disabled. It also has to do with that I like software to be easy to maintain, and having conditional help texts is not super fun to maintain; it quickly gets out of hand as the number of optional features grow :)

Btw, I plan to make a semi-big update to the "how to diff" part of the README, like update screenshots, and make `diff latest` more prominent (because I think that is one of the main use of the tool, at least it has turned out to be the case for me).

Before I do that update with screenshots, I think I would like https://github.com/Enselic/cargo-public-api/issues/251 to be fixed. Just out of curiosity, are you still working on solving that one? If not, maybe I can take a shot at it.

I'll (squash) merge this PR after I have resolved the (trivial) merge conflicts with main.

Thanks a lot for your input on this PR so far :pray:

PR #265, opened 2023-01-02
Fix rendering of bounds for Generic Associated Types (GATs)

Enselic:


Enselic:
I don't this PR will be controversial, so going ahead with merging.

PR #268, opened 2023-01-05
Replace `sub_prefix` with `ImplKind::Inherent`

Enselic:
And also rename `Normal` to `Trait` so that we use the same terminology as in https://doc.rust-lang.org/reference/items/implementations.html

> Implementation : InherentImpl | TraitImpl

This is preparation for https://github.com/Enselic/cargo-public-api/issues/251 that I think makes sense to do regardless of how we solve https://github.com/Enselic/cargo-public-api/issues/251

Since I'm touching code you wrote Emil, it would be great if you could take a look.

(I recommend reviewing each commit individually)


Emilgardis (approved):
lgtm!

PR #269, opened 2023-01-05
Group `impl`s better, make sorting more lexicographic

Enselic:
By rendering sortable path the same for both inherent and trait impls.
This changes sort order quite a bit, because the sort order becomes more lexicographic.

Note how this PR makes progress on #251. For example, instead of

```
impl core::convert::From<cargo_manifest::error::Error> for rustdoc_json::BuildError
impl core::convert::From<cargo_metadata::errors::Error> for rustdoc_json::BuildError
impl core::convert::From<std::io::error::Error> for rustdoc_json::BuildError
pub fn rustdoc_json::BuildError::from(source: cargo_manifest::error::Error) -> Self
pub fn rustdoc_json::BuildError::from(source: cargo_metadata::errors::Error) -> Self
pub fn rustdoc_json::BuildError::from(source: std::io::error::Error) -> Self
```

we now get

```
impl core::convert::From<cargo_manifest::error::Error> for rustdoc_json::BuildError
pub fn rustdoc_json::BuildError::from(source: cargo_manifest::error::Error) -> Self
impl core::convert::From<cargo_metadata::errors::Error> for rustdoc_json::BuildError
pub fn rustdoc_json::BuildError::from(source: cargo_metadata::errors::Error) -> Self
impl core::convert::From<std::io::error::Error> for rustdoc_json::BuildError
pub fn rustdoc_json::BuildError::from(source: std::io::error::Error) -> Self
```


Enselic:
(Had to rebase to trigger CI for PR, since the base branch was not originally `main`, and looks like GitHub does not re-evaluate when the base branch changes. https://github.com/Enselic/cargo-public-api/commit/476885c2941ec81a02903c6d73c4b65e677f7334 should prevent this problem from happening again)

Emilgardis (approved):
nice!

PR #270, opened 2023-01-05
Properly group multiple inherent `impl`s

### LINKED ISSUE #251: Group items even if multiple inherent `impl`s are used

Enselic:
"properly" meaning "group items like rustdoc HTML does it", which means each inherent `impl` is presented separately (i.e. no de-duplication)

Closes #251

Depends on #269

### DIFF
--- a/public-api/src/intermediate_public_item.rs
+++ b/public-api/src/intermediate_public_item.rs
@@ -42,6 +42,46 @@ impl<'c> NameableItem<'c> {
             sortable_name.push_str(&crate::tokens::tokens_to_string(
                 &context.render_impl(impl_, &[]),
             ));
+
+            // If this is an inherent impl, additionally add the concatenated
+            // names of all associated items to the "name" of the impl. This makes
+            // multiple inherent impls group together, even if they have the
+            // same "name".
+            //
+            // For example, consider this code:
+            //
+            //   pub struct MultipleInherentImpls;
+            //
+            //   impl MultipleInherentImpls {
+            //       pub fn impl_one() {}
+            //   }
+            //
+            //   impl MultipleInherentImpls {
+            //       pub fn impl_two() {}
+            //   }
+            //
+            // In this case, we want to group the two impls together. So
+            // the name of the first impl should be
+            //
+            //   impl MultipleInherentImpls-impl_one
+            //
+            // and the second one
+            //
+            //   impl MultipleInherentImpls-impl_two
+            //
+            if impl_.trait_.is_none() {
+                let mut assoc_item_names: Vec<&str> = impl_
+                    .items
+                    .iter()
+                    .filter_map(|id| context.crate_.index.get(id))
+                    .filter_map(|item| item.name.as_ref())
+                    .map(String::as_str)
+                    .collect();
+                assoc_item_names.sort_unstable();
+
+                sortable_name.push('-');
+                sortable_name.push_str(&assoc_item_names.join("-"));
+            }
         }
 
         sortable_name
--- a/public-api/src/public_item.rs
+++ b/public-api/src/public_item.rs
@@ -1,5 +1,6 @@
 use std::cmp::Ordering;
 use std::fmt::Display;
+use std::hash::Hash;
 
 use crate::intermediate_public_item::IntermediatePublicItem;
 use crate::render::RenderingContext;
@@ -16,7 +17,7 @@ pub(crate) type PublicItemPath = Vec<String>;
 /// of the public API of a crate. Implements [`Display`] so it can be printed. It
 /// also implements [`Ord`], but how items are ordered are not stable yet, and
 /// will change in later versions.
-#[derive(Clone, Eq, PartialEq, Hash)]
+#[derive(Clone)]
 pub struct PublicItem {
     /// Read [`crate::item_processor::sorting_prefix()`] docs for more info
     pub(crate) sortable_path: PublicItemPath,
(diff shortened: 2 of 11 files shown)


Enselic:
Had to rebase to trigger CI, but we should not get this problem in the future (see https://github.com/Enselic/cargo-public-api/pull/269#issuecomment-1374435468)

Emilgardis:
Does moving a function from one impl to another trigger a change with `cargo public-api diff --deny changes` ?

Enselic:
Good question! I wasn't sure at first, so I wrote a test for it, which I have now pushed. Turns out there is no diff for this case. IMHO that is correct since the public API has not changed.

The reason there is no diff triggered is because the diffing algo is based on HashBags. So the sorted order of items does not matter. (The first version of the algorithm depended on sort order however, but about half a year ago the diff algo was rewritten to not depend on sort order.)

Enselic:
Wait, I forgot to make the `fn`s `pub`, and when I fix that there is a diff that shows up :scream: Need to investigate a bit more here...

Enselic:
Since the "impl name" of the moved item changes, a diff is triggered. I think we need to somehow differentiate between a "sortable path" and a "ID path". Might require some significant conceptual changes...

Enselic:
One easy way to solve this is to make `Eq` and `Hash` only take `PublicItem::tokens` into account. When diffing, an item that *looks* the same (and from an API point of view *is* the same) will then be considered equal. In that sense, I like that solution.

The downside is that now two items might be `Ordering::Equal` but not `eq` to each other. I'm unsure if this is a problem. It is clearly stated that `Eq` and `Hash` needs to match, but I see no such thing for `Ord`.

So maybe it is fine to do it like that? If yes, the PR in its current state is production ready.

But if `Ord` should/must be in sync with `Eq`, I think we should make `Ord` consistent with `Eq`/`Hash` and instead use a special key/method when sorting. I'm unsure what the best way forward is here.

Emilgardis:
What do you mean with `Cmp`?

Enselic:
Sorry, I meant `Ord` (which has a `cmp` method)

Emilgardis:
If you mean `PartialOrd::partial_cmp` and by extension `Ord::cmp` then yes, it's a violation of

> 1. a == b if and only if partial_cmp(a, b) == Some(Equal).

Enselic:
Ah right, [there](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html) it is, thanks

I pushed a commit now that should make things well again

Emilgardis:
yes, my bad I could've linked it!

public-api/src/intermediate_public_item.rs
```diff
@@ -42,6 +42,46 @@ impl<'c> NameableItem<'c> {
             sortable_name.push_str(&crate::tokens::tokens_to_string(
                 &context.render_impl(impl_, &[]),
             ));
+
+            // If this is an inherent impl, additionally add the concatenated
+            // names of all associated items to "name" of the impl. This makes
```
Emilgardis: possible typo?, `to the "name"`

public-api/src/intermediate_public_item.rs
```diff
@@ -42,6 +42,46 @@ impl<'c> NameableItem<'c> {
...
+            //
+            // In this case, we want to group the two impls together. So make
+            // the name of the first impl to be
+            //
+            //   impl MultipleInherentImpls-impl_one
+            //
+            // and the second one to be
+            //
+            //   impl MultipleInherentImpls-impl_two
+            //
```
Emilgardis:
```suggestion
            // In this case, we want to group the two impls together. So
            // the name of the first impl should be
            //
            //   impl MultipleInherentImpls-impl_one
            //
            // and the second one
            //
            //   impl MultipleInherentImpls-impl_two
            //
```

public-api/src/public_item.rs
```diff
@@ -40,6 +41,34 @@ impl PublicItem {
     pub fn tokens(&self) -> impl Iterator<Item = &Token> {
         self.tokens.iter()
     }
+
+    /// Special version of `cmp` that is used to sort public items in a way that
```
Emilgardis:
```suggestion
    /// Special version of [`cmp`](Ord::cmp) that is used to sort public items in a way that
```

Emilgardis (approved):
lgtm! some minor changes/suggestions

public-api/src/intermediate_public_item.rs
Enselic: fixed!

Emilgardis:
feel free to squash without adding me as co-author :) or merge/squash merge as is :D

Enselic:
The commit history ended up too messy for my taste, so I force-push a better commit structure (where each commit passes CI). No code changes were made with the force-push. Soon merging, thanks :)

PR #271, opened 2023-01-07
CI: Start running `cargo deny check`, mostly to checks licences of deps

Enselic:
Note that we allow MPL-2.0 for one of the dependencies. See https://github.com/frewsxcv/rust-crates-index/pull/90 and the PR diff for motivation

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -23,3 +23,7 @@ exceptions = [
     #    https://github.com/frewsxcv/rust-crates-index/pull/90
     { name = "smartstring", allow = ["MPL-2.0"] },
 ]
+
+[bans]
+multiple-versions = "deny"
+wildcards = "deny"
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,25 @@
+[licenses]
+allow = [
+    "Apache-2.0",
+    "BSD-3-Clause",
+    "MIT",
+    "Unicode-DFS-2016",
+]
+
+unlicensed = "deny"
+copyleft = "deny"
+default = "deny"
+
+exceptions = [
+    # Allowed because:
+    #  * MPL-2.0 is a weak/non-viral copyleft license that doesn't cause much
+    #    trouble in practice
+    #  * Only affects the `cargo-public-api` binary and not any of our libraries
+    #    (`public-api`, `rustdoc-json`, `rustup-toolchain`). The dependency
+    #    chain is: `cargo-public-api` -> `crates-index` -> `smartstring`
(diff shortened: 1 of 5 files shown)


deny.toml
```diff
@@ -0,0 +1,25 @@
...
+    #  * MPL-2.0 is a weak/non-viral copyleft license that doesn't cause much
+    #    trouble in practice
+    #  * Only affects the `cargo-public-api` binary and not any of our libraries
+    #    (`public-api`, `rustdoc-json`, `rustup-toolchain`). The dependency
+    #    chain is: `cargo-public-api` -> `crates-index` -> `smartstring`
+    #  * Can be opted out from by disabling the `diff-latest` feature
+    #  * Upstream is hesitant to remove the dependency because of better
+    #    performance than alternatives:
+    #    https://github.com/frewsxcv/rust-crates-index/pull/90
+    { name = "smartstring", allow = ["MPL-2.0"] },
```
Enselic, line 24: @repi Hi Johan, can you confirm please that it is not a problem for Embark that there is one dependency under MPL-2.0 for `cargo-public-api`? MPL-2.0 is not mentioned in https://gist.github.com/repi/d98bf9c202ec567fd67ef9e31152f43f.

But I think there is no problem, because the copyleft of MPL-2.0 only applies to the `smartstring` dependency itself and not the `cargo-public-api` binary as a whole.

And, of course, `cargo-public-api` is not distributed by Embark, and so the obligations of the license does not even apply.

I have also noticed that you make use of this dependency yourself in your own (open source, but still) projects: https://github.com/EmbarkStudios/cargo-deny/blob/main/Cargo.lock#L1614-L1624.

So, I think this fine, but just wanted to make sure.

(If Embark is OK with this, then I imagine most other companies also are OK with this)

repi: yes this is fine! we use a few MPL-2.0 dependencies, just have to be aware to disclose changes to those crates if we fork/change them. but that is pretty easy as we do public forks of all crates we change.

we do not have it enabled by default in our license list though so indeed add exceptions for them to make it a bit more clear, just like you did here!

Emilgardis:
psst, don't forget https://github.com/Enselic/cargo-public-api/pull/274#issuecomment-1374889687 ;)

deny.toml
Enselic: Great, thanks for the confirmation!

Enselic:
> psst, don't forget [#274 (comment)](https://github.com/Enselic/cargo-public-api/pull/274#issuecomment-1374889687) ;)

Fixed! Will merge soon.

PR #275, opened 2023-01-08
migrate to node16 actions

Emilgardis:


Emilgardis:
Success!

image

.github/workflows/Release-cargo-public-api.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
       # OK, both public-api and cargo-public-api has been published under the
       # same version. Create a GitHub Release now.
-      - uses: softprops/action-gh-release@1e07f4398721186383de40550babbdf2b84acfc5 # v0.1.14
+      - uses: softprops/action-gh-release@d4e8205d7e959a9107da6396278b2f1f07af0f9b # 2022-12-09
```
Enselic, line 64: Since this has access to the crates.io token I want to be extra careful. IIRC I checked the code of the pinned revision to make sure it had nothing malicious.

Did you do the same for the new pinned version? If not, would be good. If it looks good to you then it's good to me.

.github/workflows/CI.yml
```diff
@@ -52,11 +52,11 @@ jobs:
...
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@e645b0cf01249a964ec099494d38d2da0f0b349f
         with:
           toolchain: nightly
-          profile: minimal
+      - run: rustup default stable
```
Enselic, line 59: Seems a bit inconvenient to have to undo the change of default, so I'm curious what the upsides of switching are?

Enselic (approved):
Thanks! Had a couple of comments/questions, but no blockers. Feel free to merge. I don't want to hold you back :)

.github/workflows/CI.yml
Emilgardis: We need this since the script assumes that the default is stable, eg when testing that the tool can be run with the stable toolchain, so that it switches to nightly.

without this, the test would be incomplete

.github/workflows/Release-cargo-public-api.yml
Emilgardis: no! but did now, seems fine https://github.com/softprops/action-gh-release/compare/1e07f4398721186383de40550babbdf2b84acfc5...d4e8205d7e959a9107da6396278b2f1f07af0f9b

.github/workflows/CI.yml
Enselic: Sorry for being unclear, I meant "what are the upsides of switching to dtolnay/rust-toolchain?" :)

Emilgardis: see https://github.com/actions-rs/toolchain/issues/216 and [the reddit thread](https://old.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/)

we could also simply use `rustup toolchain install nightly --profile minimal` but `dtolnay/rust-toolchain` has a nice way to specify what version to use, but we don't use it here

.github/workflows/Release-cargo-public-api.yml
Enselic: Watch out! You need to review the "diff" of `dist/index.js`! (https://github.com/softprops/action-gh-release/blob/master/action.yml#L62) The files with the nice diff are not necessarily used by the built index.js.

So it's quite a lot of work to review :grimacing:

I'm starting to wonder if I really did a review of the old pinned SHA1... 

The crates.io token we use does not allow to remove owners, so it is not a super disaster if it leaks... (But it would allow attackers to upload malicious `cargo-public-api` versions, and since there (AFAIK) is no way to get email notifications when new versions of a crate are published, it could cause some real problems for a while)

.github/workflows/CI.yml
Enselic: Ooooh I didn't know the other one was abandoned. Then this is definitely the right move :+1:

.github/workflows/Release-cargo-public-api.yml
Emilgardis: did a diff of the minified js and what my build returns

```diff
diff --git a/dist/index.js b/dist/index.js
index 0728eb3..ba6bf15 100644
--- a/dist/index.js
+++ b/dist/index.js
@@ -12374,46 +12374,52 @@
       a.upload = upload;
       const release = (e, p, d = 3) =>
         t(void 0, void 0, void 0, function* () {
-          var t, r;
+          var t, r, o, n;
           if (d <= 0) {
             console.log(`❌ Too many retries. Aborting...`);
             throw new Error("Too many retries.");
           }
-          const [o, n] = e.github_repository.split("/");
-          const l =
+          const [l, c] = e.github_repository.split("/");
+          const u =
             e.input_tag_name ||
             ((0, s.isTag)(e.github_ref)
               ? e.github_ref.replace("refs/tags/", "")
               : "");
-          const c = e.input_discussion_category_name;
-          const u = e.input_generate_release_notes;
+          const v = e.input_discussion_category_name;
+          const h = e.input_generate_release_notes;
           try {
             if (e.input_draft) {
               try {
                 for (
-                  var v = i(p.allReleases({ owner: o, repo: n })), h;
-                  (h = yield v.next()), !h.done;
+                  var g = true, w = i(p.allReleases({ owner: l, repo: c })), b;
+                  (b = yield w.next()), (t = b.done), !t;
 
                 ) {
-                  const e = h.value;
-                  let a = e.data.find((e) => e.tag_name === l);
-                  if (a) {
-                    return a;
+                  n = b.value;
+                  g = false;
+                  try {
+                    const e = n;
+                    let a = e.data.find((e) => e.tag_name === u);
+                    if (a) {
+                      return a;
+                    }
+                  } finally {
+                    g = true;
                   }
                 }
               } catch (e) {
-                t = { error: e };
+                r = { error: e };
               } finally {
                 try {
-                  if (h && !h.done && (r = v.return)) yield r.call(v);
+                  if (!g && !t && (o = w.return)) yield o.call(w);
                 } finally {
-                  if (t) throw t.error;
+                  if (r) throw r.error;
                 }
               }
             }
-            let a = yield p.getReleaseByTag({ owner: o, repo: n, tag: l });
+            let a = yield p.getReleaseByTag({ owner: l, repo: c, tag: u });
             const d = a.data.id;
-            let g;
+            let _;
             if (
               e.input_target_commitish &&
               e.input_target_commitish !== a.data.target_commitish
@@ -12421,47 +12427,47 @@
               console.log(
                 `Updating commit from "${a.data.target_commitish}" to "${e.input_target_commitish}"`
               );
-              g = e.input_target_commitish;
+              _ = e.input_target_commitish;
             } else {
-              g = a.data.target_commitish;
-            }
-            const w = l;
-            const b = e.input_name || a.data.name || l;
-            const _ = (0, s.releaseBody)(e) || "";
-            const T = a.data.body || "";
-            let E;
-            if (e.input_append_body && _ && T) {
-              E = T + "\n" + _;
+              _ = a.data.target_commitish;
+            }
+            const T = u;
+            const E = e.input_name || a.data.name || u;
+            const y = (0, s.releaseBody)(e) || "";
+            const x = a.data.body || "";
+            let S;
+            if (e.input_append_body && y && x) {
+              S = x + "\n" + y;
             } else {
-              E = _ || T;
+              S = y || x;
             }
-            const y =
+            const k =
               e.input_draft !== undefined ? e.input_draft : a.data.draft;
-            const x =
+            const D =
               e.input_prerelease !== undefined
                 ? e.input_prerelease
                 : a.data.prerelease;
-            const S = yield p.updateRelease({
-              owner: o,
-              repo: n,
+            const O = yield p.updateRelease({
+              owner: l,
+              repo: c,
               release_id: d,
-              tag_name: w,
-              target_commitish: g,
-              name: b,
-              body: E,
-              draft: y,
-              prerelease: x,
-              discussion_category_name: c,
-              generate_release_notes: u,
+              tag_name: T,
+              target_commitish: _,
+              name: E,
+              body: S,
+              draft: k,
+              prerelease: D,
+              discussion_category_name: v,
+              generate_release_notes: h,
             });
-            return S.data;
+            return O.data;
           } catch (t) {
             if (t.status === 404) {
-              const t = l;
-              const i = e.input_name || l;
+              const t = u;
+              const i = e.input_name || u;
               const r = (0, s.releaseBody)(e);
-              const v = e.input_draft;
-              const h = e.input_prerelease;
+              const o = e.input_draft;
+              const n = e.input_prerelease;
               const g = e.input_target_commitish;
               let w = "";
               if (g) {
@@ -12470,16 +12476,16 @@
               console.log(`👩‍🏭 Creating new GitHub release for tag ${t}${w}...`);
               try {
                 let e = yield p.createRelease({
-                  owner: o,
-                  repo: n,
+                  owner: l,
+                  repo: c,
                   tag_name: t,
                   name: i,
                   body: r,
-                  draft: v,
-                  prerelease: h,
+                  draft: o,
+                  prerelease: n,
                   target_commitish: g,
-                  discussion_category_name: c,
-                  generate_release_notes: u,
+                  discussion_category_name: v,
+                  generate_release_notes: h,
                 });
                 return e.data;
               } catch (t) {
```
looks good

Enselic: :star:

PR #276, opened 2023-01-09
CI: Use plain `rustup install` instead of `dtolnay/rust-toolchain`

Enselic:
@Emilgardis As you pointed out, we can use plain `rustup install`. After thinking some more, I think I prefer this. The motivation goes something like this:

> The less we depend on third party code, the easier it is to understand our own systems. Not only for ourselves, but also for people new to the project. And the lower the barrier, the higher the chance of getting contributions. In addition, third party code always increases the attack surfaces for malicious actors.

I appreciate the capability of `dtolnay/rust-toolchain` to specify `toolchain: stable minus 8 releases`, but I don't see us needing that. Having a generous MSRV takes time to maintain, and IMHO it is not worth it.

If you strongly prefer us to keep using `dtolnay/rust-toolchain` then I am fine with that too. Feel free to close this PR in that case.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -53,10 +53,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@e645b0cf01249a964ec099494d38d2da0f0b349f
-        with:
-          toolchain: nightly
-      - run: rustup default stable
+      - run: rustup install --profile minimal nightly
       - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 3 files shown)


Emilgardis:
no objections!

One thing we could add though is a problem matcher but that's a bit overkill imo, see https://github.com/cross-rs/cross/blob/main/.github/actions/setup-rust/rust.json and https://github.com/cross-rs/cross/blob/8c86a98d119b63ad624728bd5e18cae23fefdac2/.github/actions/setup-rust/action.yml#L26 for an example

Enselic:
Cool stuff, but yeah probably overkill at this point I think :)

PR #281, opened 2023-01-10
CI-EXAMPLES.md: Only recommend the `cargo test` approach

Enselic:
It is much better than using the CLI in special CI jobs.

### DIFF
--- a/docs/CI-EXAMPLES.md
+++ b/docs/CI-EXAMPLES.md
@@ -4,57 +4,54 @@ This document describes different ways to make use of `cargo public-api` in CI. 
 
 ## Prevent Accidental Public API Changes
 
-### With a Public API Set in Stone
+### With Blessable `cargo test`
 
-If the API is set in stone, you can use the `--deny=all` flag together with `diff ...` to deny all kinds of changes (additions, changes, removals) to your public API. A GitHub Actions job to do this for PRs would look something like this:
+The best way to version the public API of your crate - and require any changes to show up in diffs - is to write a regular `cargo test` that you run via CI along with all other tests.
 
-```yaml
-jobs:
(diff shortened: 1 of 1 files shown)


docs/CI-EXAMPLES.md
```diff
@@ -4,57 +4,54 @@ This document describes different ways to make use of `cargo public-api` in CI.
...
-```
+This create a `tests/public-api.txt` file in your project that you version together with your other project files. Any changes to it (and the public API) will show up in e.g. PR diffs.
 
-#### Locking
+### Locking
 
-Since the rustdoc JSON format is unstable and frequently changes across nightly toolchain versions, and since improvements to `cargo public-api` are regularly released, you probably want to lock against a specific version of `cargo public-api` and a specific version of the nightly toolchain. To find matching versions, consult the [Compatibility Matrix](../README.md#compatibility-matrix). Then use the syntax above to provision CI with these versions.
+Since the rustdoc JSON format is unstable and frequently changes across nightly toolchain versions, and since improvements to `cargo public-api` are regularly released, you probably want to lock against a specific version of `public-api` and a specific version of the nightly toolchain. The above example code does that. To find other matching versions, consult the [Compatibility Matrix](../README.md#compatibility-matrix).
```
Emilgardis, line 57: nit, no it doesn't, to be exact you'd have to use `=x.y.z`, in `Cargo.toml` `version = "0.1.0"` is equivalent to semver `"^0.1.0"`

Enselic: Good that you brought that up! Because I had in mind an update to the versioning guidelines, I just forgot to write it down. So I created a PR for it now: https://github.com/Enselic/cargo-public-api/pull/286

PR #283, opened 2023-01-11
add logging of commands

Emilgardis:
image


rustdoc-json/src/ext.rs
```diff
@@ -0,0 +1,45 @@
...
+        let args = self.get_args().map(std::ffi::OsStr::to_string_lossy);
+        let program = self.get_program().to_string_lossy();
+        let args = shell_words::join(args);
+        let space = if args.is_empty() { "" } else { " " };
+        if let Some(cwd) = self.get_current_dir() {
+            log::debug!("running `{} {program}{space}{args}`", cwd.display(),);
+        } else {
+            log::debug!("running `{program}{space}{args}`",);
+        }
+    }
```
Emilgardis, line 44: unfortunately, the track_caller does nothing

https://github.com/rust-lang/log/pull/410

clippy.toml
```diff
@@ -0,0 +1,6 @@
+# enable as needed
+disallowed-methods = [
+#    { path = "std::process::Command::status", reason = "use the extension CommandExt::status_and_log defined in this crate instead" },
+#    { path = "std::process::Command::output", reason = "use the extension CommandExt::output_and_log defined in this crate instead" },
+#    { path = "std::process::Command::spawn", reason = "use the extension CommandExt::spawn_and_log defined in this crate instead" },
+]
```
Emilgardis, line 6: used this mostly to just find all usages, doesn't need to be included

rustdoc-json/src/ext.rs
Emilgardis: I'd definitely opt into using tracing simply for this reason

cargo-public-api/Cargo.toml
```diff
@@ -12,10 +12,11 @@ license = "MIT"
 repository = "https://github.com/Enselic/cargo-public-api"
 
 [features]
-default = ["diff-latest"]
+default = ["diff-latest", "log"]
```
Enselic, line 15: How many more deps needs to be built with `log` enabled? If it is say 3 or less I would prefer to build with logging support unconditionally (i.e. not have a `log` as an opt-out feature). If it is say 10 or more additional deps, I think we should make it an non-default (opt-in) feature.

I don't think we should have it as an opt-out feature. It is debatable if we even need to have `diff-latest` as a feature.

Each feature not only doubles the number of build combinations (2^n) but also puts burden on users to make a choice. I like to think of lack of cargo `features` as us doing users a favor.

But as always, I am very open to hear counter-arguments :)

clippy.toml
```diff
@@ -0,0 +1,6 @@
+# enable as needed
+disallowed-methods = [
```
Enselic, line 2: I think it makes sense to always disallow these, going forward, right? To make sure we keep logging also in new code.

clippy.toml
```diff
@@ -0,0 +1,6 @@
...
+#    { path = "std::process::Command::status", reason = "use the extension CommandExt::status_and_log defined in this crate instead" },
+#    { path = "std::process::Command::output", reason = "use the extension CommandExt::output_and_log defined in this crate instead" },
```
Enselic, line 4: Can these be put in `scripts/cargo-clippy.sh`? So that there is just one clippy "config". Or maybe get rid of that script and put it all in `clippy.toml` (not sure if that is doable/desirable)

rustdoc-json/Cargo.toml
```diff
@@ -9,12 +9,17 @@ readme = "README.md"
...
+[features]
+log = ["dep:log", "dep:shell-words"]
+
 [dependencies]
 cargo_metadata = "0.15.2"
 cargo-manifest = "0.7.1"
 serde = { version = "1.0.152", features = ["derive"] }
 thiserror = "1.0.38"
 toml = "0.5.10"
+log = { version = "0.4.17", optional = true }
```
Enselic, line 21: I've played with the thought of introducing good logging before, and had planned on using `tracing`, since that is what `rustc` uses. But maybe `log` & friends is a better choice? It probably is, since that was what you picked :)

rustdoc-json/src/ext.rs
```diff
@@ -0,0 +1,45 @@
+#![doc(hidden)]
+#![allow(clippy::disallowed_methods, clippy::module_name_repetitions)]
```
Enselic, line 2: Isn't it better to scope `allow(clippy::disallowed_methods` only to the statements/expressions that need it, rather than doing it for the whole module? That way it is less risk of introducing code that SHOULD be disallowed, but that is not detected.

rustdoc-json/src/ext.rs
Enselic: Could you explain a bit more please? I don't quite understand. I'm pretty sure I understand what `track_caller` is supposed to do, but not how it applies to the logging.

Enselic (commented):
Looks good on first sight! And also seems very useful. Had some comments/questions.

cargo-public-api/Cargo.toml
Emilgardis: 0 :)

edit: sorry, 1 with env_logger, 2 with shell_words, log was already included in the crate graph, and all env_logger deps too

rustdoc-json/Cargo.toml
Emilgardis: No I picked log since I know you don't want to add many dependencies, but my choice is definitely `tracing`, i'll explain more [here](https://github.com/Enselic/cargo-public-api/pull/283#pullrequestreview-1244320510)

rustdoc-json/src/ext.rs
Emilgardis: the linked pr makes it so that crates like `env_logger` show [`Record::file`](https://docs.rs/log/0.4.17/log/struct.Record.html#method.file) and `line` using `std::panic::Location`, meaning if the log call is burried deep (like here, behind two functions) it'll bubble up to the thing that called `status_and_log`

i.e the log would, with how this pr is right now (if `env_logger` had a way to enable file:line), be


```
[DEBUG src/builder.rs:18]
```

instead of how it shows now

```
[DEBUG src/ext.rs:40]
```

Emilgardis: `tracing` already has this functionality, and definitely has better presets for log output, but, many of the features are overkill and are unfortunately not possible to opt out of.

clippy.toml
Emilgardis: unfortunately no, `disallowed-methods` is strictly a `clippy.toml` thing, and `clippy.toml` can only take [these lints](https://rust-lang.github.io/rust-clippy/master/index.html#Configuration)

see https://github.com/rust-lang/rust-clippy#configure-the-behavior-of-some-lints

Enselic:
@Emilgardis Would you be OK with closing this PR until you have time to resume work on it? (But there is no obligation of course.) 

I like when the PR inbox is empty 😎

PR #287, opened 2023-01-12
Fix CI with Rust 1.66 clippy

Enselic:


Enselic:
Looks like GitHub runners also suffers from this problem... I bet if we just wait a few days this will be sorted out upstream

CC @Emilgardis 

(I need to stop for today)

(Also, I am playing around with auto-merge. I hope I did not mess things up too much. Feel free to tinker.)

Emilgardis (approved):
lgtm

Emilgardis:
the required statuses are broken

imo, the best way to have required statuses is to make a simple job that just depends/needs the other jobs, that way you'll minimize the requirements

Emilgardis:
So something like

```yaml
  conclusion:
    runs-on: ubuntu-latest
    needs: [cargo-fmt, cargo-doc, cargo-clippy, cargo-audit, cargo-deny, cargo-test]
    steps:
      - run: exit 0
```

and then the required status would be `CI / conclusion`

Enselic:
> So something like
> 
> ```yaml
>   conclusion:
>     runs-on: ubuntu-latest
>     needs: [cargo-fmt, cargo-doc, cargo-clippy, cargo-audit, cargo-deny, cargo-test]
>     steps:
>       - run: exit 0
> ```
> 
> and then the required status would be `CI / conclusion`

That seems like a good way of doing it, thanks. I think I'll set that up later. I removed the required statuses again for now.

PR #285, opened 2023-01-11
[INVALID] correctly add features when using cache

Emilgardis:
resolves #282

PR #289, opened 2023-01-13
Support for selecting features with `cargo public-api diff x.y.z`

### LINKED ISSUE #282: cannot enable features for crate with diff
```bash
$ git clone https://github.com/Enselic/cargo-public-api.git
$ cd cargo-public-api
$ cargo run -p cargo-public-api -- --all-features -p public-api diff latest
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/cargo-public-api --all-features --package public-api diff latest`
Resolved `diff latest` to `diff 0.26.0`
error: cannot specify features for packages outside of workspace
Error: Failed to build rustdoc JSON. Stderr: See above
```

Emilgardis:
resolves #282

continuation of #285

### DIFF
--- a/cargo-public-api/src/published_crate.rs
+++ b/cargo-public-api/src/published_crate.rs
@@ -31,7 +31,18 @@ pub fn build_rustdoc_json(version: impl Into<String>, args: &Args) -> Result<Pat
     };
 
     write_file("lib.rs", "// empty lib")?;
-    let manifest = write_file("Cargo.toml", &manifest_for(&spec))?;
+    let (manifest, needs_resolved_package) = manifest_simple(args, &spec)?;
+    let manifest = write_file("Cargo.toml", &manifest)?;
+
+    if needs_resolved_package {
+        let mut metadata = cargo_metadata::MetadataCommand::new();
+        metadata.manifest_path(&manifest);
+        let metadata = metadata.exec()?;
+        if let Some(package) = metadata.packages.iter().find(|p| p.name == spec.name) {
+            // XXX if metadata doesn't find a package, rustdoc will tell us why, so we can continue
+            write_file("Cargo.toml", &manifest_with_info(args, &spec, package)?)?;
+        };
+    }
 
     // Since we used `crate::builder_from_args(args)` above it means that if
     // `args.target_dir` is set, both the dummy crate and the real crate will
@@ -39,7 +50,10 @@ pub fn build_rustdoc_json(version: impl Into<String>, args: &Args) -> Result<Pat
     // won't work. So always clear the target dir before we use the builder.
     let builder = crate::api_source::builder_from_args(args)
         .clear_target_dir()
-        .manifest_path(manifest)
+        .all_features(false)
+        .features(Vec::<&str>::new())
+        .no_default_features(false)
+        .manifest_path(&manifest)
         .package(&spec.name);
     crate::api_source::build_rustdoc_json(builder)
 }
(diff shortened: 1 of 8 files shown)


cargo-public-api/src/published_crate.rs
```diff
@@ -31,15 +31,33 @@ pub fn build_rustdoc_json(version: impl Into<String>, args: &Args) -> Result<Pat
...
+            metadata.manifest_path(&manifest);
+            let metadata = metadata.exec()?;
+            let Some(package) = metadata.packages.iter().find(|p| p.name == spec.name) else {
+                // XXX if metadata doesn't find a package, rustdoc will tell us why, so continue
+                break 'resolve;
+            };
+
+            write_file("Cargo.toml", &manifest_with_info(args, &spec, package)?)?;
+        }
+    }
```
Emilgardis: Looks reasonable on a high level, thanks! Playing around with the code a bit, I noticed that even if I remove this whole `if needs_resolved` block, all tests still pass. Would be great if we could come up with a way to test the "needs_resolved" part. Otherwise I will for sure break it by accident in the future :sweat_smile:

_Originally posted by @Enselic in https://github.com/Enselic/cargo-public-api/pull/285#discussion_r1067686451_

Emilgardis: I'm unable to make a test for this without making a new version of example_api featuring features

Enselic: Sound good! I think that will become useful on more occasions in the future. I'll add you as an owner on crates.io when I get home this evening!

cargo-public-api/src/published_crate.rs
```diff
@@ -198,3 +210,19 @@ impl PackageSpec {
...
+        &Args::try_parse_from(["test", "-p", "example-api", "diff", "0.1.1"].iter()).unwrap(),
+        &PackageSpec {
+            name: "example-api".to_owned(),
+            version: "0.1.1".to_owned(),
+        },
+    )
+    .unwrap()
+    .0;
+    assert_eq!(manifest, "[package]\nname = \"crate-downloader\"\nversion = \"0.1.0\"\nedition = \"2021\"\n\n[lib]\npath = \"lib.rs\"\n\n[dependencies.example-api]\nversion = \"=0.1.1\"\n");
+}
```
Emilgardis: accidental regression in this pr, added a test for it :)

cargo-public-api/src/published_crate.rs
```diff
@@ -114,13 +114,23 @@ fn is_true(b: &bool) -> bool {
...
-struct Dep<'a, S: AsRef<str>> {
+struct Dependency<'a, S: AsRef<str>> {
+    #[serde(serialize_with = "add_equal")]
     version: &'a str,
     #[serde(skip_serializing_if = "is_true")]
     default_features: bool,
     #[serde(skip_serializing_if = "<[_]>::is_empty")]
     features: &'a [S],
+    #[serde(skip_serializing_if = "Option::is_none")]
+    registry: Option<&'a str>,
```
Emilgardis: used this for using a custom registry instead of pushing to crates.io. We should probably actually support overriding the registry, but that's not in the scope of this pr, should be removed

Enselic:
Is this PR ready for another review round?

Emilgardis:
Yes!

cargo-public-api/src/published_crate.rs
```diff
@@ -31,15 +31,33 @@ pub fn build_rustdoc_json(version: impl Into<String>, args: &Args) -> Result<Pat
     };
 
     write_file("lib.rs", "// empty lib")?;
-    let manifest = write_file("Cargo.toml", &manifest_for(&spec))?;
+    let (manifest, needs_resolved) = dbg!(manifest_simple(args, &spec)?);
```
Enselic: Should `dbg!` still be here?

cargo-public-api/src/published_crate.rs
```diff
@@ -31,15 +31,33 @@ pub fn build_rustdoc_json(version: impl Into<String>, args: &Args) -> Result<Pat
...
+    let manifest = write_file("Cargo.toml", dbg!(&manifest))?;
+
+    'resolve: {
+        if needs_resolved {
```
Enselic: Even after reading the code in some detail I still can't make sense semantically of `needs_resolved`. How should that term be interpreted :thinking: ?

cargo-public-api/src/published_crate.rs
```diff
@@ -31,15 +31,33 @@ pub fn build_rustdoc_json(version: impl Into<String>, args: &Args) -> Result<Pat
     };
 
     write_file("lib.rs", "// empty lib")?;
-    let manifest = write_file("Cargo.toml", &manifest_for(&spec))?;
+    let (manifest, needs_resolved) = dbg!(manifest_simple(args, &spec)?);
+    let manifest = write_file("Cargo.toml", dbg!(&manifest))?;
+
+    'resolve: {
```
Enselic: Nit (feel free to ignore): I find it hard to get a good grasp of the structure of the code when labels are used. Is there really no way to use regular looping constructs here?

cargo-public-api/src/published_crate.rs
```diff
@@ -31,15 +31,33 @@ pub fn build_rustdoc_json(version: impl Into<String>, args: &Args) -> Result<Pat
...
     // `args.target_dir` is set, both the dummy crate and the real crate will
     // write to the same JSON path since they have the same project name! That
     // won't work. So always clear the target dir before we use the builder.
     let builder = crate::api_source::builder_from_args(args)
         .clear_target_dir()
-        .manifest_path(manifest)
+        .all_features(false)
+        .features(Vec::<&str>::new())
+        .no_default_features(false)
+        .manifest_path(&manifest)
```
Enselic, line 56: Nit (feel free to ignore): Maybe the code ends up nicer if we instead of "copy all and clear some" instead used "copy only the ones we want to copy"?

cargo-public-api/src/published_crate.rs
```diff
@@ -91,20 +109,94 @@ fn build_dir(args: &Args, spec: &PackageSpec) -> PathBuf {
...
+
+fn add_equal<S>(ver: &str, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    serializer.serialize_str(&format!("={ver}"))
+}
+
+#[derive(serde::Serialize)]
+struct Dependency<'a, S: AsRef<str>> {
```
Enselic: Can't we use https://docs.rs/cargo-manifest/0.7.1/cargo_manifest/enum.Dependency.html instead of defining our own version of this?

Enselic (commented):
Looks good overall! The most important thing is the regression test. Thank you for writing one.

Had a handful of comments. Feel free to ignore the nits.

cargo-public-api/src/published_crate.rs
Emilgardis: Unfortunately with the current api it's not possible, can't access the fields of a Builder

cargo-public-api/src/published_crate.rs
Emilgardis: done! I didn't use it because it does useless allocations, but it's so minor it doesn't matter

cargo-public-api/src/published_crate.rs
Emilgardis: removed the label, replaced with a `if let`

cargo-public-api/src/published_crate.rs
Emilgardis: renamed it to `needs_resolved_package`, does that signal the intent better?

Enselic (approved):
LGTM!

PR #292, opened 2023-01-14
Change `rustdoc_json::Builder::toolchain(...)` to take `Into<String>`

Enselic:
Also add `clear_toolchain()`. (We already have `clear_target_dir()`.)

This makes the `fn public_api` test in `README.md` look nicer.

Don't bother to go through a deprecation cycle because that adds too much complexity and overhead in this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ fn public_api() {
 
     // Build rustdoc JSON
     let rustdoc_json = rustdoc_json::Builder::default()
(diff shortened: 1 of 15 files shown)


Enselic:
Merging since I have a bunch of follow-up commits, and I don't want to build up a giant tower of dependent PRs. Feel free to make retroactive code review comments, and I'll take care of them.

PR #293, opened 2023-01-14
Replace `PublicApi::from_rustdoc_json(path, options)` with `public_api::Builder`

Enselic:
Purpose of this draft PR: To announce my intention to replace the current `PublicApi::from_rustdoc_json(path, options)` API with a `public_api::Builder` based API.

Benefits of doing this API change:
* Nicer API
* More symmetry with the `rustdoc_json::Builder` API

 TODO:
* Keep prototyping a bit more
* Figure out deprecation/migration strategy for clients
* Split up into smaller increments that are easier to review
* Should we have `public_api::Builder::default().rustdoc_json(path)` instead of `public_api::Builder::from_rustdoc_json(path)` perhaps?

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,9 +107,9 @@ fn public_api() {
         .unwrap();
 
     // Derive the public API from the rustdoc JSON
-    let public_api =
-        public_api::PublicApi::from_rustdoc_json(rustdoc_json, public_api::Options::default())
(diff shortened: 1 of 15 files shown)


Enselic:
Unsure when I will have time to do this. Closing this draft PR for now.

PR #294, opened 2023-01-15
Support `cargo public-api diff highest` (highest = highest semver version)

Enselic:
This also fixes CI since 0.3.0 is not latest any longer, now 0.2.1 is latest. I'm confused how scheduled nightly didn't fail, because locally the test fails for me.

CC @Emilgardis  (no worries :)

Since CI is broken I will merge this as soon as CI pass in this PR.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,7 +9,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  CARGO_NET_GIT_FETCH_WITH_CLI: true # Workaround from https://github.com/rust-lang/cargo/issues/11014
 
 jobs:
   cargo-fmt:
--- a/.github/workflows/Nightly.yml
+++ b/.github/workflows/Nightly.yml
@@ -7,7 +7,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  CARGO_NET_GIT_FETCH_WITH_CLI: true # Workaround from https://github.com/rust-lang/cargo/issues/11014
 
 jobs:
   # The purpose of this job is to detect problems like
--- a/.github/workflows/Release-cargo-public-api.yml
+++ b/.github/workflows/Release-cargo-public-api.yml
@@ -5,7 +5,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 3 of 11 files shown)


Enselic:
What the heck, this fails only on Ubuntu. I have no idea why, and no time right now to investigate deeper.

Enselic:
I got strange errors like this locally:
```
stderr="Resolved `diff latest` to `diff 0.2.1`\n    Updating crates.io index\nerror: failed to get `example_api` as a dependency of package `crate-downloader v0.1.0 (/tmp/.tmpVtuky9/cargo-public-api/build-root-for-published-crates/example_api-0.2.1)`\n\nCaused by:\n  failed to load source for dependency `example_api`\n\nCaused by:\n  Unable to update registry `crates-io`\n\nCaused by:\n  failed to fetch `https://github.com/rust-lang/crates.io-index`\n\nCaused by:\n  process didn\'t exit successfully: `git fetch --force --update-head-ok \'https://github.com/rust-lang/crates.io-index\' \'+HEAD:refs/remotes/origin/HEAD\'` (exit status: 128)\n  --- stderr\n  fatal: unable to access \'https://github.com/rust-lang/crates.io-index/\': Could not resolve host: github.com\nError: Failed to build rustdoc JSON. Stderr: See above\n"
```
and figured maybe this has to do with the workaround I pushed a few days ago.

And after reverting the workaround code, now CI is green again :raised_hands: 

Merging.

Emilgardis:
Hmm, I think latest should be highest, e.g in cargo terms `latest = { version = "*" }`

Enselic:
That might very well be the case, but where is that terminology used? Skimming https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html it does not seem used there.

Emilgardis:
I don't think there is a spelled out definition of what `*` (and latest) means, but see for example https://github.com/rust-lang/cargo/issues/10741

Enselic:
I think "most recently published" might indeed be a bit odd and not get much usage. So I changed `diff latest` to mean "diff highest published semver version" and removed `diff highest` again. See https://github.com/Enselic/cargo-public-api/pull/296

PR #296, opened 2023-01-15
Make `diff latest` diff highest semver version, not most recently published

Enselic:
And remove `diff highest` since we have not yet deployed that to users. And adding `diff recent` does not seem very useful, but will add a non-negligible maintenance cost.


Emilgardis (approved):
lgtm!

PR #303, opened 2023-01-25
Bump `crates-index` to `0.19.0`; `0.18.12` is yanked

Enselic:
Unfortunately we also need to allow multiple version in Cargo.lock for now because of this.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -25,5 +25,5 @@ exceptions = [
 ]
 
 [bans]
-multiple-versions = "deny"
+# multiple-versions = "deny"
 wildcards = "deny"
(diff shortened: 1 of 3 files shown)


deny.toml
```diff
@@ -25,5 +25,5 @@ exceptions = [
 ]
 
 [bans]
-multiple-versions = "deny"
+# multiple-versions = "deny"
 wildcards = "deny"
```
Emilgardis, line 29: You can allow certain packages to have dupes.

https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-skip-field-optional

So here we'd do

```toml
[[bans.skip]]
name = "rand"
version = "<0.19"
```

i think

Enselic: Thanks, I was in a bit of a rush this morning and didn't have time to look this up, but didn't want to leave CI red either

Fixed in https://github.com/Enselic/cargo-public-api/pull/304

@Emilgardis BTW, purely out of curiosity: Do you also get an email when Nightly CI fails?

Emilgardis: No, I've turned it off 

[image](https://github.com/settings/notifications)

PR #353, opened 2023-03-18
Introduce `public_api::Builder`, deprecate `PublicApi::from_rustdoc_json()` and `Options`

### LINKED ISSUE #318: Change `PublicApi::from_rustdoc_json(path, Options::default())` to `public_api::Builder::from_rustdoc_json(path).build()`
Instead of this public API for the `public-api` lib:

```rust
    // Derive the public API from the rustdoc JSON
    let mut options = public_api::Options::default();
    options.omit_blanket_impls = true;
    options.omit_auto_trait_impls = true;
    let public_api = public_api::PublicApi::from_rustdoc_json(path, options).unwrap();
```

I want the public API to be this:

```rust
    // Derive the public API from the rustdoc JSON
    let public_api = public_api::Builder::from_rustdoc_json(path)
        .omit_blanket_impls(true)
        .omit_auto_trait_impls(true)
        .build()
        .unwrap();
```

which has the following benefits:
* nicer and simpler API
* symmetry with the `rustdoc_json::Builder` API

I have made a prototype commit that passed CI: https://github.com/Enselic/cargo-public-api/pull/293

Doing it properly with a proper public API deprecation path is more work, however. And this is not top prio for me right now. So I am creating this issue to make my long-term intention visible to others.

Enselic:
Instead of this public API:

    // Derive the public API from the rustdoc JSON
    let mut options = public_api::Options::default();
    options.omit_blanket_impls = true;
    options.omit_auto_trait_impls = true;
    let public_api = public_api::PublicApi::from_rustdoc_json(path, options).unwrap();

The public API is now this:

    // Derive the public API from the rustdoc JSON
    let public_api = public_api::Builder::from_rustdoc_json(path)
        .omit_blanket_impls(true)
        .omit_auto_trait_impls(true)
        .build()
        .unwrap();

which has the following benefits:
 * nicer and simpler API
 * symmetry with the `rustdoc_json::Builder` API

Closes #318

### DIFF
--- a/public-api/src/lib.rs
+++ b/public-api/src/lib.rs
@@ -5,20 +5,12 @@
 //! If you want a convenient CLI for this library, you should use [cargo
 //! public-api](https://github.com/Enselic/cargo-public-api).
 //!
-//! As input to the library, a special output format from `cargo doc` is used,
-//! which goes by the name **rustdoc JSON**. Currently, only `cargo doc` from
-//! the Nightly toolchain can produce **rustdoc JSON** for a library. You build
-//! **rustdoc JSON** like this:
+//! As input to the library, a special output format from `rustdoc +nightly` is
+//! used, which goes by the name **rustdoc JSON**. Currently, only the nightly
+//! toolchain can build **rustdoc JSON**.
 //!
-//! ```bash
-//! cargo +nightly rustdoc --lib -- -Z unstable-options --output-format json
-//! ```
-//!
-//! Consider using [`rustdoc_json`](https://crates.io/crates/rustdoc_json)
-//! instead of invoking the above command yourself.
-//!
-//! The main entry point to the library is [`PublicApi::from_rustdoc_json_str`],
-//! so please read its documentation.
+//! You use the [`rustdoc-json`](https://crates.io/crates/rustdoc_json) library
+//! to programmatically build rustdoc JSON. See below for example code.
 //!
 //! # Examples
 //!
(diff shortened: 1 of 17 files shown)


public-api/src/lib.rs
```diff
@@ -74,25 +62,66 @@ pub const MINIMUM_NIGHTLY_VERSION: &str = "nightly-2023-01-04";
...
 #[derive(Copy, Clone, Debug)]
-#[non_exhaustive] // More options are likely to be added in the future
 #[allow(clippy::struct_excessive_bools)]
-pub struct Options {
+struct BuilderOptions {
+    sorted: bool,
+    debug_sorting: bool,
+    omit_blanket_impls: bool,
+    omit_auto_trait_impls: bool,
+    omit_auto_derived_impls: bool,
```
Emilgardis, line 73: why not `pub`?

Enselic: We want the public API surface to be minimal, and I don't think we need these pub?

Enselic: (Note that BuilderOptions is not pub. I plan on renaming BuilderOptions to Options once we remove the old pub Options)

Emilgardis: only concern is that there's no way to check what it's been set to, but it's not a big problem

Enselic: If a use case shows up where that is needed we can certainly consider adding some API for it, but in that case I think we want getters on `Builder`

PR #360, opened 2023-03-20
Try again to fix the all-jobs job

Enselic:
We must use `always()`, because the default `success()` will evaluate to
`skipped` if a needed matrix job fails, and a required status check for a
job is considered OK if it is skipped. And since we use `always()`, we
must check the result of each job manually.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,7 +12,7 @@ env:
 
 jobs:
   all-jobs:
-    if: always()
+    if: always() # Otherwise this job is skipped if the matrix job fails
     name: all-jobs
     runs-on: ubuntu-latest
     needs:
@@ -23,8 +23,7 @@ jobs:
       - cargo-deny
       - cargo-test
     steps:
-      - run: |
-          [ "${{ needs.cargo-test.result }}" = "success" ]
+      - run: jq --exit-status 'all(.result == "success")' <<< '${{ toJson(needs) }}'
 
   cargo-fmt:
     name: cargo fmt -- --check


.github/workflows/CI.yml
```diff
@@ -24,7 +28,13 @@ jobs:
...
     steps:
       - run: |
-          [ "${{ needs.cargo-test.result }}" = "success" ]
+          [ "${{ needs.cargo-fmt.result }}" = "success" ] &&
+          [ "${{ needs.cargo-doc.result }}" = "success" ] &&
+          [ "${{ needs.cargo-clippy.result }}" = "success" ] &&
+          [ "${{ needs.cargo-audit.result }}" = "success" ] &&
+          [ "${{ needs.cargo-deny.result }}" = "success" ] &&
+          [ "${{ needs.cargo-test.result }}" = "success" ] &&
+          echo "All dependent jobs were successful"
```
Emilgardis:
```suggestion
          jq -r 'all(.result == "success")' <<< "${needs}" &&
          echo "All dependent jobs were successful"
```

Enselic: @Emilgardis Would be great if that worked, but I don't think `"${needs}"` expands to anything? 🤔 

```
$  jq -r 'all(.result == "success")' <<< "${empty}" && echo yes
yes
```

Here is a test-job where I print its value, but it seems empty: https://github.com/git-branch-deleter/git-branch-deleter/actions/runs/4472287212/jobs/7858262225

```
$ Run cat <<< "${needs}"

```

Emilgardis: sorry, missing a crucial part :D

```yml
        env:
          needs: ${{ toJson(needs) }}
```

this is basically stolen from cross' ci [here](https://github.com/cross-rs/cross/blob/99b8069c0d977a14cd421ad8a3ef3255dc5802be/.github/workflows/ci.yml#L361)

Enselic: Aha, it is `toJson()` that is the secret sauce. Few seems to know about this trick, because I could not find it via Google. Thanks a lot for this very nice tip :)

PR #362, opened 2023-03-22
Introduce `rustup_toolchain::Installer` to make API future proof and configurable

Enselic:
I'm a bit unsure of exactly how the API should look though. In the PR it looks like this:

```rust
    rustup_toolchain::Installer::default()
        .toolchain(toolchain)
        .run()
        .unwrap();
```

but maybe it should look like this:

```rust
    rustup_toolchain::Installer::of(toolchain)
        .run()
        .unwrap();
```

but I like the way `Installer::default()` communicates "I am using a sane set of defaults for you and you can use this as-is if you don't want any other changes". On the other hand, very few is going to want to use the default `--profile minimal stable` toolchain installer...

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,10 @@ Then add the following test to your project. As the author of the below test cod
 #[test]
 fn public_api() {
     // Install a proper nightly toolchain if it is missing
-    rustup_toolchain::ensure_installed(public_api::MINIMUM_NIGHTLY_VERSION).unwrap();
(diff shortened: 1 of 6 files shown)


Enselic:
When I wrote this I felt sure I wanted a more generic API.

But now I have starte to feel like this is just adding unneecessary complexity.

Maybe in the future we can do something like this.

For now, let's just not.

PR #365, opened 2023-03-25
rustup-toolchain: Rename `ensure_installed()` to `install()` and deprecate `is_installed()`

Enselic:
Rename `ensure_installed()` to `install()` for brevity. Deprecate `is_installed()` to make it clear it is not needed around `install()`.


Enselic:
I want to get a release out today of all crates, so I will merge this right away.

Normally I try to do wait a day or two before I merge significant PRs.

PR #367, opened 2023-03-25
Interpret `cargo public-api diff` as `cargo public-api diff latest`

Enselic:


Emilgardis (approved):
this makes perfect sense

PR #385, opened 2023-04-01
Remove diffing from `public-api` bin (NOT from `cargo public-api`)

Enselic:
I never use the `public-api` bin, and I suspect nobody else does it either. If they do, they should switch to `cargo public-api`.

Remove diffing support to test the hypothesis that nobody uses it. If no one complains, it indicates that nobody uses it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -734,6 +734,7 @@ dependencies = [
  "expect-test",
  "hashbag",
  "itertools",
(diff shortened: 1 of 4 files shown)


Emilgardis (approved):
I'll complain, this broke my workflow...

kidding, feel free to merge!

Enselic:
😀

PR #386, opened 2023-04-01
Remove the `public-api` bin (NOT the `cargo-public-api` bin)

Enselic:
The `public-api` bin is a simple CLI wrapper around the `public_api` library. It comes nowhere near the capabilities of `cargo public-api`. So no one uses it. It had a role to play in the early development of this project, but it's not needed anymore.

So let's remove it. It reduces `cargo test` time and maintenance overhead.

The `cargo test`s that are removed in this commit are already covered by other tests elsewhere in the project, or specific to the `public-api` bin.


Emilgardis (approved):
lets do it!

Enselic:
🥳

PR #406, opened 2023-04-26
Use `[features]` info from crates index

Enselic:
We don't need to resolve feature info ourselves with `cargo_metadata::MetadataCommand`, because the info we need is already present in the crates index.


Emilgardis (approved):
awesome!

PR #416, opened 2023-05-03
RELEASE.md: Use `cargo-edit` for bumping

Enselic:
It is getting tiresome to manually bump version in all places. This is an attempt to automate it.

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -5,9 +5,7 @@
 1. First release `rustdoc-json` and `rustup-toolchain` if needed. See below.
 1. Update `public-api/CHANGELOG.md`
 1. Create a PR that targets `main` that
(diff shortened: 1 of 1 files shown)


Emilgardis:
have you seen cargo-edit and its [`cargo upgrade`](https://github.com/killercup/cargo-edit#cargo-upgrade)?

Emilgardis:
ooh! I misunderstood the problem this solves... I thought it was normal dependencies but that's already solved with dependabot.

cargo-edit has another thing for this though, [`cargo set-version -p <package> <ver>`](https://github.com/killercup/cargo-edit#cargo-set-version)

that works just fine

Enselic:
Nice! I have heard about cargo-edit, but I didn't know it had these capabilities. It indeed seems to work just fine.

PR #419, opened 2023-05-09
Rustdoc-JSON: Replace `-` with `_` for binary package target

frondeus:
Hi! First of all, thanks for this very useful crate :)

Whenever I want to build JSON documentation for a binary with `-` in it, I get a `PathBuf` that does not point to the generated file.

This PR is fixing that and adding a new test that ensures correctness.

I am opening this as a draft because perhaps a similar solution should also be applied to other package target kinds.


rustdoc-json/src/builder.rs
```diff
@@ -144,8 +144,8 @@ fn rustdoc_json_path_for_manifest_path(
 
             lib_name.replace('-', "_")
         }
-        PackageTarget::Bin(package)
-        | PackageTarget::Example(package)
+        PackageTarget::Bin(package) => package.replace('-', "_"),
+        PackageTarget::Example(package)
         | PackageTarget::Test(package)
```
Enselic, line 139: Probably we should always replace - with _, regardless of PackageTarget? I think we can assume that to be the case, so please replace - with _ for all PackageTargets. You do not need to add any additional regression tests for that.

rustdoc-json/tests/test_crates/test_crate/Cargo.toml
```diff
@@ -8,6 +8,11 @@ edition = "2021"
 test_dep_1 = { path = "../test_dep_1_0_0", package = "test_dep" }
 test_dep_2 = { path = "../test_dep_2_0_0", package = "test_dep" }
 
+
+[[bin]]
+name = "with-hypen"
+path = "src/bin/with_hypen.rs"
```
Enselic: If you rename the file to with-hypen.rs I think you don't need this [[bin]] section at all? Not 100% sure though. But it would be nice to not have it if we don't have to.

Enselic (changes requested):
Thanks a lot for taking the time to open a PR, and especially thank you for the regression test 👍 

Glad you like the project :)

Just curious: What is your use case? In what scenario do you want to list the public API for a bin?

frondeus:
> Just curious: What is your use case? In what scenario do you want to list the public API for a bin?

I am only using the rustdoc-JSON and writing a library that would provide a compile-time reflection based on the JSON output.

Enselic:
@frondeus There is no urgency of course; just wanted to check if you intend to bring this PR to the finish line?

frondeus:
Yes I do, as soon as I return from my holiday leave :)

Enselic (approved):

Enselic:
I'd like to make a release this weekend, so I've helped you take this to the finish line. I hope you didn't mind!

frondeus:
Absolutely! Thanks for your work :)

Enselic:
Fixed in https://crates.io/crates/rustdoc-json/0.8.6

PR #430, opened 2023-05-30
Ignore `!` when sorting `impl`s to make `Send` and `Sync` order stable

### LINKED ISSUE #429: Generic type information missing for items in paths
Hi, I was trying to use cargo public-api diffs on skia-safe, which uses generic handle types for a lot of the exported types. Here is an minimum working example that explains the problem:

lib.rs:
```rust
pub struct Handle<T>(T);

pub type HU32 = Handle<u32>;

impl HU32 {
    pub fn get_u32() -> u32 {
        0
    }
}

```

results in

```rust
...
pub mod public_api_test
pub struct public_api_test::Handle<T>(_)
impl public_api_test::Handle<u32>
pub fn public_api_test::Handle::get_u32() -> u32
...
pub type public_api_test::HU32 = public_api_test::Handle<u32>
```

but I I think it the functions should be represented with their complete type signature:

```rust
...
pub fn public_api_test::Handle<u32>::get_u32() -> u32
```

Without that, a diff compares all functions of the generic type instead of only the concrete type.

Enselic:
We especially do this for auto traits so that they remain in the same order no matter if some of them are negative impls or not.

If there are no objections, we should still wait with merging this, because we need to bump 0.x.0 when we release this: https://github.com/Enselic/cargo-public-api/blob/main/docs/MAINTAINER.md#versioning-strategy. And I'd like to also fix #429 as part of next 0.x.0 bump.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,10 @@ exclude = [
     # https://stackoverflow.com/questions/56713877/why-do-proc-macros-have-to-be-defined-in-proc-macro-crate
     "test-apis/comprehensive_api_proc_macro",
 
+    # Tests that requires auto traits. Normally we omit auto trait impls because
+    # they make the output very repetitive.
+    "test-apis/auto_traits",
+
     # Contains different versions of the same small API. Used mainly to test
     # test the API-diffing functionality of this library.
     "test-apis/example_api-v0.1.0",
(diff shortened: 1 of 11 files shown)


Enselic:
I feel close enough to a solution for #429 to be comfortable with merging this

Issue #408, opened 2023-04-27
Support sparse creates.io registry


Milo123459:
When enabling sparse registry, I consistently got this error:
```
Error: The repo at path /home/codespace/.cargo/registry/index/github.com-1ecc6299db9ec823 is unusable due to having an invalid HEAD reference: reference 'refs/heads/main' not found; class=Reference (4); code=NotFound (-3)
```
Once disabled, and recreating the registry (rming it then cargo fetch) it works.

Command: `cargo public-api diff latest`

Enselic:
Thanks a lot for reporting. I can reproduce with our git repo like this. Note how I get a different error first:
```
$ export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse

$ rm -rf ~/.cargo/registry

$ cargo run -- -p rustup-toolchain diff latest
[...]
Error: there is no TLS stream available; class=Ssl (16)
Caused by:
    there is no TLS stream available; class=Ssl (16)

$ cargo run -- -p rustup-toolchain diff latest
[...]
Error: The repo at path /home/martin/.cargo/registry/index/github.com-1ecc6299db9ec823 is unusable due to having an invalid HEAD reference: reference 'refs/heads/master' not found; class=Reference (4); code=NotFound (-3)
```

Looks like we either need to wait for https://github.com/frewsxcv/rust-crates-index/issues/87 or roll our own solution. Personally I would wait, but if someone comes up with a nice and simple way of doing it ourselves, I would be open to merge that until our deps support it.  But honestly it would be much better to put that effort into adding support upstream.

Milo123459:
I also saw that error first, then after a bit, it went away and switched to the one I sent in the issue.

Enselic:
The registry is only used when diffing. More specifically, it is used to create a "fake" Cargo project so that the rustdoc JSON for an old version can be built. I wonder if maybe we should always use the "sparse" index to get what we need for creating the "fake" Cargo.toml, and then let cargo use whatever it wants. So maybe we can do something like this:

```console
$ curl https://index.crates.io/ex/am/example_api
{"name":"example_api","vers":"0.1.0","deps":[],"cksum":"d0f99c425515721620183a74993ebb809ca7f06c4e6b09b8c28668d56a2e78fd","features":{},"yanked":false}
[...]

PR #438, opened 2023-06-07
CI: Set CARGO_REGISTRIES_CRATES_IO_PROTOCOL=git until we support sparse

Enselic:
See #408


Enselic:
Our nightly failed because of this this night: https://github.com/Enselic/cargo-public-api/actions/runs/5195717032/jobs/9368708518

However, now it doesn't fail anymore. Maybe they reverted that change because they noticed CI started to fail for a lot of people.

I will wait with submitting this until we see the problem at least once more.

PR #454, opened 2023-07-01
Stop enforcing pedantic clippy

Enselic:
Over time I have ended up finding it more annoying than helpful to have pedantic clippy. In other words, I have come over to your side Emil (see https://github.com/Enselic/cargo-public-api/pull/117#pullrequestreview-1087455759). (I'm even skeptical about enabling specific pedantic lints now.)


Emilgardis (approved):
Hehehe, very good

PR #481, opened 2023-08-15
Bless `nightly-2023-08-15` output

EnselicCICD:
Automatically created by https://github.com/Enselic/cargo-public-api/actions/runs/5863327157/jobs/auto-bless


Enselic:
Seems caused by https://github.com/rust-lang/rust/pull/113464 which is fine

PR #485, opened 2023-08-22
Remove $ from README commands to ease copy-paste

tim-day-387:
Thanks for maintaining such a cool tool!

I noticed that the Github copy/paste widget erroneously picks up these $ (and some other text). I have a commit that removes everything besides the commands in these command snippets to make it a little quicker to install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,14 +6,14 @@ List and diff the public API of Rust library crates between releases and commits
 
 Install the `cargo public-api` subcommand with a recent regular **stable** Rust toolchain:
 
-```console
-$ cargo +stable install cargo-public-api --locked
+```sh
+cargo +stable install cargo-public-api --locked
 ```
 
 Ensure **nightly-2023-05-24** or later is installed (does not need to be the active toolchain) so `cargo public-api` can build rustdoc JSON for you:
 
-```console
-$ rustup install nightly --profile minimal
+```sh
+rustup install nightly --profile minimal
 ```
 
 # Usage
@@ -22,14 +22,14 @@ $ rustup install nightly --profile minimal
 
 This example lists the public API of the `regex` crate. First we clone the repo:
 
-```console
-$ git clone https://github.com/rust-lang/regex ; cd regex
+```sh
+git clone https://github.com/rust-lang/regex ; cd regex
 ```
 
 Now we can list the public API of `regex` by running
 
-```console
-$ cargo public-api
+```sh
+cargo public-api
 ```
 
 which will print the public API of `regex` with one line per public item in the API:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,13 +7,13 @@ List and diff the public API of Rust library crates between releases and commits
 Install the `cargo public-api` subcommand with a recent regular **stable** Rust toolchain:
 
 ```console
```
Enselic: Let's change from `console` to `sh` here and everywhere now that the line is valid `sh`.

README.md
```diff
@@ -52,16 +52,13 @@ $ cargo public-api diff 1.6.0
 ### … Against the Latest Published Version
 
 ```console
-$ cargo public-api diff latest
-Resolved `diff latest` to `diff 1.7.1`
-[...]
```
Enselic: I'd like to keep the output in this case because I find it quite informative. For consistency maybe I should create a screenshot here too. Until then, maybe we can separate the command from its output like this? Or is that too confusing? Note how the first block is `sh` but the second is just pure text.

```sh
cargo public-api diff latest
```

```
Resolved `diff latest` to `diff 1.7.1`
[...]
```

README.md
```diff
@@ -52,16 +52,13 @@ $ cargo public-api diff 1.6.0
...
+cargo public-api diff latest
 ```
 
 ### … Between Git Refs
 
 ```console
-$ cargo public-api diff ref1..ref2
-[...]
```
Enselic, line 64: Yes, it's probably OK to remove `[...]` here.

README.md
```diff
@@ -132,9 +129,13 @@ are included in the list of public items by default. Use
...
-$ cargo public-api --omit blanket-impls,auto-trait-impls,auto-derived-impls
+cargo public-api --omit blanket-impls,auto-trait-impls,auto-derived-impls
+```
+
+Or a shorter form of the above command:
 
-$ cargo public-api -sss    # Shorter form of above command
+```console
+cargo public-api -sss
 ```
```
Enselic: Good idea to move the comment out as proper text 👍 I think we can reduce repetition a bit by rearranging the text like this:

    respectively to omit such items from the output to make it much less noisy:
    
    ```sh
    cargo public-api --omit blanket-impls,auto-trait-impls,auto-derived-impls
    ```
    
    For convenience you can also use `-s` (`--simplified`) to achieve the same thing. This is a shorter form of the above command:
    
    ```sh
    cargo public-api -sss
    ```

Enselic (changes requested):
Happy to hear you like it, and thank you for taking the time to make a contribution! I agree it would be nice if the commands were conveniently copy-pasteable. I just had some comments on the specifics.

README.md
tim-day-387: Keeping `console` for the second part renders that text as blue, which should make it a bit clearer.

Enselic (approved):

Enselic:
Looks good, thanks!

PR #489, opened 2023-08-26
rustdoc-json: Add `rustdoc_json::Builder::verbose(bool)`

Enselic:


Emilgardis (commented):
I don't see what this is supposed to help with

Enselic:
My use case was that I got `SIGILL` (see https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/SIGILL.20in.20build-script-build.20with.20nightly-2023-08-25)

I got `SIGILL` when building rustdoc JSON. I wanted to see the command to see if I could reproduce the SIGILL with just the rustdoc command. To help me come up with a MCVE.

So I had one use for it. And I'm thinking one use probably means more uses in the future?

Enselic:
@Emilgardis I don't want to merge this if you are against it, so I'm curious on if you could elaborate your concerns a bit? Is it the API surface you are concerned about? Or something else?

Emilgardis (approved):
alright, lets do it, my only concern was that it didn't seem useful, but maybe it could be and it doesn't do any harm

PR #511, opened 2023-10-31
Bless `nightly-2023-10-31` output

EnselicCICD:
Automatically created by https://github.com/Enselic/cargo-public-api/actions/runs/6701734965/jobs/auto-bless


Enselic:
Looks like https://github.com/rust-lang/rust/pull/112463

PR #513, opened 2023-11-13
Add context when running command fails

### LINKED ISSUE #512: Make error message more helpful when `rustup` is not in `PATH`
I just cloned a fresh repository and executed `cargo public-api` on it.

Its output was `Error: No such file or directory (os error 2)` which is not very descriptive of what went wrong.

Please consider this a bugreport.

matthiasbeyer:
Closes #512 

Please have a look.
This does not remove the transparent IO error (yet), but its a good start I guess.

### DIFF
--- a/rustdoc-json/src/lib.rs
+++ b/rustdoc-json/src/lib.rs
@@ -46,6 +46,17 @@ pub enum BuildError {
     #[error(transparent)]
     CargoMetadataError(#[from] cargo_metadata::Error),
 
+    /// Cannot run a command dues to std::io::Error
+    #[error("Cannot run command {:?}", .command.get_program())]
+    CannotRunCommand {
+        /// The command that failed
+        command: std::process::Command,
+
+        /// The underlying IO error
+        #[source]
+        source: std::io::Error,
+    },
+
     /// Some kind of IO error occurred.
     #[error(transparent)]
     IoError(#[from] std::io::Error),
--- a/rustdoc-json/src/builder.rs
+++ b/rustdoc-json/src/builder.rs
@@ -17,7 +17,14 @@ pub fn run_cargo_rustdoc(options: Builder) -> Result<PathBuf, BuildError> {
     if options.verbose {
         eprintln!("Running: {:?}", cmd);
     }
-    if cmd.status()?.success() {
+    let command_result = cmd
+        .status()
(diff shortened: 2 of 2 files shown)


matthiasbeyer:
Lets see what CI tells us... I cannot link this (have to investigate what's missing in my setup, nix things, don't mind that please :laughing: )

Emilgardis:
did this fix your error message also?

rustdoc-json/src/lib.rs
```diff
@@ -46,6 +46,17 @@ pub enum BuildError {
...
+        command: std::process::Command,
+
+        /// The underlying IO error
+        #[source]
+        source: std::io::Error,
+    },
+
     /// Some kind of IO error occurred.
     #[error(transparent)]
     IoError(#[from] std::io::Error),
```
Emilgardis, line 62: this can be removed now I think since that command call is the only usage

Emilgardis (changes requested):

rustdoc-json/src/lib.rs
Emilgardis: that's a breaking change though, so maybe just mark as deprecated

Enselic: personally I would prefer to keep it, it is very generic and likely to be needed in the future, and I'm not sure it is worth the effort to try to deprecate it

rustdoc-json/src/lib.rs
```diff
@@ -46,6 +46,17 @@ pub enum BuildError {
     #[error(transparent)]
     CargoMetadataError(#[from] cargo_metadata::Error),
 
+    /// Cannot run a command dues to std::io::Error
+    #[error("Cannot run command {:?}", .command.get_program())]
+    CannotRunCommand {
```
Enselic, line 51: One reason CI is failing is because you need to bless this new public API

Running `./scripts/bless-expected-output-for-tests.sh` should work

rustdoc-json/src/lib.rs
Enselic: ...or maybe we should always convert to more descriptive error messages.... but I still think it's fine to keep around for now

Enselic:
Closing for now to keep the PR inbox clean. Of course feel free to re-open this if you resume work on it!

PR #520, opened 2023-12-01
Include all subcommands in top-level `--help` output

### LINKED ISSUE #238: Plain `--help` should also show `diff --help`
So that a single `--help` invocation is sufficient to learn how to use the tool.

Blocked on https://github.com/clap-rs/clap/issues/1334 (unless we work around this somehow, which I would be in favor doing, unless it's very complicated).

Enselic:
So that users only have to issue one command to see how all subcommands are supposed to be used.

Closes #238

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -352,7 +352,7 @@ fn main_task_from_diff_args(args: &Args, diff_args: &DiffArgs) -> Result<MainTas
         )
     }
 
-    let first_arg = diff_args.args.get(0);
+    let first_arg = diff_args.args.first();
     let second_arg = diff_args.args.get(1);
 
     let main_task = match (first_arg, second_arg) {
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -31,6 +31,7 @@ mod vendor;
     long_about = "List and diff the public API of Rust library crates between releases and commits. Website: https://github.com/Enselic/cargo-public-api",
     bin_name = "cargo public-api"
 )]
+#[command(flatten_help = true)]
 pub struct Args {
     /// Path to `Cargo.toml`.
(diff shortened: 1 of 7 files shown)


Enselic:
Actually, let me see if I can update the help text with examples also with flatten_help...

Edit: Done.

cargo-public-api/src/main.rs
```diff
@@ -145,8 +146,25 @@ struct DiffArgs {
...
 
-    /// What to diff. See `cargo public-api diff --help` for examples and more
-    /// info.
+    #[clap(verbatim_doc_comment)]
+    /// What to diff.
+    ///
+    /// EXAMPLES
+    /// ========
+    ///
+    /// Diffing working tree against a published version of a specific crate in the workspace:
```
Emilgardis:
```suggestion
    /// Diffing current working tree against a published version of `specific-crate` in the workspace:
```

cargo-public-api/src/main.rs
```diff
@@ -145,8 +146,25 @@ struct DiffArgs {
...
+    ///
+    ///     cargo public-api -p specific-crate diff 1.2.3
+    ///
+    /// Diffing between commits:
+    ///
+    ///     cargo public-api diff 1.2.3 cargo public-api diff v0.2.0..v0.3.0
```
Emilgardis:
```suggestion
    ///     cargo public-api diff v0.2.0..v0.3.0
```

Emilgardis (changes requested):

cargo-public-api/src/main.rs
Enselic: Oops, thanks :)

cargo-public-api/src/main.rs
Enselic: 👍 

Inspired by your suggestion I reworded it a bit further to make it shorter so it fits on a single 100 char line.

Emilgardis (approved):

PR #531, opened 2024-01-28
Bless `nightly-2024-01-28` output

EnselicCICD:
Automatically created by https://github.com/Enselic/cargo-public-api/actions/runs/7683167097/jobs/auto-bless


Enselic:
Caused by https://github.com/rust-lang/rust/pull/116167 which is fine

PR #551, opened 2024-03-12
Bless `nightly-2024-03-12` output

EnselicCICD:
Automatically created by https://github.com/Enselic/cargo-public-api/actions/runs/8243210039/jobs/auto-bless


Enselic (approved):
https://github.com/rust-lang/rust/pull/121840

PR #555, opened 2024-03-21
Remove `cargo public-api --toolchain foo` arg. Use `cargo +foo public-api` instead.

Enselic:
Maintaining both variants is too much work. From now on we only support `cargo +foo public-api`.

But users will normally not specify a toolchain explicitly and simply run `cargo public-api` instead.

This will make it easier to solve https://github.com/Enselic/cargo-public-api/issues/512.


Enselic:
@Emilgardis Would be great if you could take a look at this. Any objections? I don't think we have a good reason to keep supporting two different ways of specifying a specific toolchain?

Emilgardis (approved):
I agree 👍

Issue #561, opened 2024-03-28
Diffing public API against a published crate with `[package] name` != `[lib] name` miscalculates the rustdoc JSON path and fails


V0ldek:
This is similar to an earlier issue I reported (and that was fixed): #476 

This time I don't have an easy repro, because it seems to require the crate to be actually published on crates.io. You could nonetheless repro it with the project I'm working on: https://github.com/V0ldek/rsonpath

## Bug

Diffing against a previous version of the crate doesn't work if used to check a library on a package with a different name.

The root cause appears to be similar to #476, only that this time the rustdoc file gets cached into `$HOME/.cache/cargo-public-api/build-root-for-published-crates`.

## Repro

1. Clone `rsonpath` and go to the library crate:

```console
> git clone https://github.com/V0ldek/rsonpath.git && cd ./rsonpath/crates/rsonpath-lib
```

**Note**: if my hypothesis is correct this should work with any package that produces a library with a different name and was previously published to crates.io.

2. Try to diff against a published version, for example 0.8.6:

```console
> cargo public-api diff 0.8.6
   Compiling proc-macro2 v1.0.79
   Compiling unicode-ident v1.0.12
   Compiling libc v0.2.153
   Compiling getrandom v0.1.16
    Checking cfg-if v1.0.0
   Compiling syn v1.0.109
    Checking ppv-lite86 v0.2.17
   Compiling thiserror v1.0.58
    Checking minimal-lexical v0.2.1
    Checking memchr v2.7.2
    Checking unicode-width v0.1.11
    Checking log v0.4.21
    Checking smallvec v1.13.2
    Checking static_assertions v1.1.0
    Checking nom v7.1.3
   Compiling quote v1.0.35
   Compiling syn v2.0.55
    Checking memmap2 v0.9.4
    Checking rand_core v0.5.1
    Checking rand_chacha v0.2.2
    Checking rand v0.7.3
   Compiling thiserror-impl v1.0.58
    Checking rsonpath-syntax v0.2.0
   Compiling contracts v0.4.0
    Checking vector-map v1.0.1
 Documenting rsonpath-lib v0.8.6
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.22s
Error: Failed to parse rustdoc JSON at "/home/mat/.cache/cargo-public-api/build-root-for-published-crates/rsonpath-lib/-/0.8.6/target/doc/rsonpath_lib.json".

This version of `cargo public-api` requires at least:

[...]

PR #564, opened 2024-03-29
Use the `tracing` crate for debug logging and e.g. `RUST_LOG=debug` to activate

Enselic:
From previous discussions my impression is that ideally we would use the `tracing` crate for logging.

While investigating https://github.com/Enselic/cargo-public-api/issues/561 I got a good reason to improve debug logging, so I switched to `tracing`. It does add a handful of more deps, but I think it's worth it. Great logging capabilities are very nice to have.

Obviously we can add more logging later, but this is a start.

I don't think anyone is using `verbose`, so going through a deprecation cycle would be overkill. Let's just remove it but release a new 0.x.0 version just to be safe.


cargo-public-api/src/main.rs
```diff
@@ -290,6 +290,7 @@ pub enum Action {
 const LATEST_VERSION_ARG: &str = "latest";
 
 fn main_() -> Result<()> {
+    tracing_subscriber::fmt::init();
```
Emilgardis: I think the default is to log to stdout, see https://togithub.com/tokio-rs/tracing/issues/2492

should be changed to stderr via the builder

Emilgardis (changes requested):
Nice! a minor note

cargo-public-api/src/main.rs
Enselic: Thanks, good catch. I have now added a regression test that would have caught this mistake (and will catch similar mistakes in the future).

Emilgardis (approved):

PR #569, opened 2024-04-27
Make most CLI options global so they also work as subcommand args (e.g. `cargo public-api diff --package=...`)

Enselic:
Before, the following would not work, because the `--package` and `--color` options were not allowed after subcommands such as `diff`

    cargo public-api diff --package foo --color always

Now that command works. The help texts becomes more noisy, but I think that is a fair price.

I added a couple of regression tests that fails without the fix, but I did not bother to be exhaustive.

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -34,11 +34,11 @@ mod vendor;
 #[command(flatten_help = true)]
 pub struct Args {
     /// Path to `Cargo.toml`.
-    #[arg(long, value_name = "PATH", default_value = "Cargo.toml")]
+    #[arg(global = true, long, value_name = "PATH", default_value = "Cargo.toml")]
     manifest_path: PathBuf,
 
     /// Name of package in workspace to list or diff the public API for.
-    #[arg(long, short)]
+    #[arg(global = true, long, short)]
     package: Option<String>,
 
     /// Omit noisy items. Can be used more than once.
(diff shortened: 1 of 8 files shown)


cargo-public-api/src/main.rs
```diff
@@ -91,31 +91,31 @@ pub struct Args {
...
     /// patch versions.
-    #[arg(long, hide = true)]
+    #[arg(global(true), long, hide = true)]
     verbose: bool,
 
     /// Show the hidden "sorting prefix" that makes items nicely grouped
     ///
     /// Only intended for debugging this tool.
-    #[arg(long, hide = true)]
+    #[arg(global(true), long, hide = true)]
```
Emilgardis: I think `global = true` should be used instead, but I don't have a very strong opinion about it, just not used to seeing it done like this with clap

Emilgardis (approved):

cargo-public-api/src/main.rs
Enselic: I stumbled across this when exploring the [code](https://github.com/rust-lang/rust/blob/a6803b9de4e7ab146b442162316b340f887a7796/src/bootstrap/src/core/config/flags.rs#L44) of bootstrap and didn't think much more about it, I just copied it.

But clearly that is not following our existing style. Fixed. Thanks for pointing it out 👍 

(Edit: upstream inconsistency fixed by https://github.com/rust-lang/rust/pull/124445)

PR #572, opened 2024-05-02
Implement JSON output

### LINKED ISSUE #570: Output JSON?
I would like to be able to postprocess my API diffs I generate with `cargo-public-api`. Would it be possible to add a feature so that `cargo-public-api` can output JSON?

matthiasbeyer:
Closes #570

### DIFF
--- a/cargo-public-api/src/json.rs
+++ b/cargo-public-api/src/json.rs
@@ -1,0 +1,71 @@
+use std::io::Write;
+
+use public_api::{diff::PublicApiDiff, PublicApi};
+
+pub struct Json;
+
+#[derive(serde::Serialize)]
+struct Api {
+    items: Vec<PublicItem>,
+}
+
+#[derive(serde::Serialize)]
+struct Diff {
+    added: Vec<PublicItem>,
+    removed: Vec<PublicItem>,
+    changed: Vec<ChangedItem>,
+}
+
+#[derive(serde::Serialize)]
+#[serde(transparent)]
+struct PublicItem(String);
+
+#[derive(serde::Serialize)]
+struct ChangedItem {
+    old: PublicItem,
+    new: PublicItem,
+}
+
+impl Json {
+    pub fn print_items(w: &mut dyn Write, api: PublicApi) -> std::io::Result<()> {
+        serde_json::to_writer(
+            w,
+            &Api {
+                items: api
+                    .items()
+                    .into_iter()
+                    .map(|itm| PublicItem(itm.to_string()))
+                    .collect(),
+            },
+        )?;
+
+        Ok(())
+    }
+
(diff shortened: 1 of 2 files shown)


cargo-public-api/src/json.rs
```diff
@@ -0,0 +1,71 @@
...
+
+pub struct Json;
+
+#[derive(serde::Serialize)]
+struct Api {
+    items: Vec<PublicItem>,
+}
+
+#[derive(serde::Serialize)]
+struct Diff {
```
Enselic, line 13: I think I would prefer us to make `PublicApiDiff` serializable directly rather than making copies of the data structures.

I haven't done a detailed review but this does not look completely unreasonable in general.

But again, I would like to see how this can be used, to better understand the use case. Maybe  there is a way to achieve your end goal without adding JSON support to cargo-public-api. Supporting a JSON format will not come for free (in terms of spare time that has to be spent maintaining it), and I would like that cost to be a bit more justified.

matthiasbeyer: > I think I would prefer us to make PublicApiDiff serializable directly rather than making copies of the data structures.

Yep I agree that would be better. I added the view structs so that I did not have to mess with the rest of the codebase, but you're totally right.

> But again, I would like to see how this can be used, to better understand the use case

What I am doing right now is this: https://github.com/matthiasbeyer/actions-to-website/blob/master/.github/workflows/branch-public-api.yml but I'd like to add a custom template for that comment I generate in there.

> Maybe there is a way to achieve your end goal without adding JSON support to cargo-public-api.

Sure, if you have an idea...?

matthiasbeyer: Sorry for not getting back to this yet, I am currently traveling and am on vacation the next two weeks. I will come back to this afterwards.

Enselic: No worries, take your time.

It's not entirely clear to me yet what exactly you want to do though. I have no idea what `nix --extra-experimental-features "nix-command flakes" develop .#pa` does for example.

matthiasbeyer: I want to render the data this tool computes with a custom template.

Enselic: I'm thinking maybe we should libbify cargo-public-api so that anyone can use its code.

And then you can do

* your program  ->
* cargo-public-api lib ->
* public-api lib ->
* PublicApiDiff 

instead of

* your program ->
* cargo-public-api bin ->
* public-api lib ->
* PublicApiDiff ->
* JSON ->
* PublicApiDiff

maybe you can even use the public-api lib directly? It depends on your exact use case.

Enselic:
Let'c close this to keep the PR inbox clean. Feel free to re-open if/when you resume work on this. We can of course keep discussing this feature.

PR #574, opened 2024-05-12
Allow `stdout` and `stderr` of rustdoc JSON building to be captured with new `rustdoc_json::Builder::build_with_captured_output(self, stdout: impl Write, stderr: impl Write)` function

orium:
Hi @Enselic. This PR allows users to capture the std{out,err} of rustdoc, which is something I need to do in cargo-rdme.

Let me know what you think.

### DIFF
--- a/rustdoc-json/src/builder.rs
+++ b/rustdoc-json/src/builder.rs
@@ -1,6 +1,7 @@
 use super::BuildError;
 use tracing::*;
 
+use std::io::Write;
 use std::{
     path::{Path, PathBuf},
     process::Command,
@@ -11,16 +12,42 @@ use std::{
 /// specify what toolchain, by temporarily changing this.
 const OVERRIDDEN_TOOLCHAIN: Option<&str> = option_env!("RUSTDOC_JSON_OVERRIDDEN_TOOLCHAIN_HACK"); // Some("nightly-2022-07-16");
 
-/// Run `cargo rustdoc` to produce rustdoc JSON and return the path to the built
-/// file.
-pub fn run_cargo_rustdoc(options: Builder) -> Result<PathBuf, BuildError> {
+struct CaptureOutput<O, E> {
+    stdout: O,
+    stderr: E,
+}
+
+fn run_cargo_rustdoc<O, E>(
+    options: Builder,
+    capture_output: Option<CaptureOutput<O, E>>,
+) -> Result<PathBuf, BuildError>
+where
+    O: Write,
+    E: Write,
+{
     let mut cmd = cargo_rustdoc_command(&options)?;
     info!("Running {cmd:?}");
-    if cmd
-        .status()
-        .map_err(|e| BuildError::General(format!("Failed to run `{cmd:?}`: {e}")))?
-        .success()
-    {
+
+    let status = match capture_output {
+        Some(CaptureOutput {
+            mut stdout,
+            mut stderr,
+        }) => {
+            let cmd_result = cmd
+                .output()
+                .map_err(|e| BuildError::General(format!("Failed to run `{cmd:?}`: {e}")))?;
+
+            stdout.write_all(&cmd_result.stdout)?;
+            stderr.write_all(&cmd_result.stderr)?;
+
+            cmd_result.status
+        }
+        None => cmd
+            .status()
+            .map_err(|e| BuildError::General(format!("Failed to run `{cmd:?}`: {e}")))?,
+    };
+
+    if status.success() {
         rustdoc_json_path_for_manifest_path(
             &options.manifest_path,
             options.package.as_deref(),
@@ -33,7 +60,7 @@ pub fn run_cargo_rustdoc(options: Builder) -> Result<PathBuf, BuildError> {
         if manifest.package.is_none() && manifest.workspace.is_some() {
             Err(BuildError::VirtualManifest(options.manifest_path))
         } else {
-            Err(BuildError::General(String::from("See above")))
+            Err(BuildError::CrateBuildError)
         }
     }
 }
@@ -51,6 +78,7 @@ fn cargo_rustdoc_command(options: &Builder) -> Result<Command, BuildError> {
         target,
         quiet,
         silent,
+        color,
         no_default_features,
         all_features,
         features,
@@ -93,6 +121,11 @@ fn cargo_rustdoc_command(options: &Builder) -> Result<Command, BuildError> {
         command.stdout(std::process::Stdio::null());
         command.stderr(std::process::Stdio::null());
     }
+    match *color {
+        Color::Always => command.arg("--color").arg("always"),
+        Color::Never => command.arg("--color").arg("never"),
+        Color::Auto => command.arg("--color").arg("auto"),
+    };
     command.arg("--manifest-path");
     command.arg(manifest_path);
     if let Some(target) = target {
@@ -214,6 +247,17 @@ fn library_name(
     Ok(package.name.clone())
 }
 
+/// Color configuration for the output of `cargo rustdoc`.
+#[derive(Clone, Copy, Debug)]
+pub enum Color {
+    /// Always output colors.
+    Always,
+    /// Never output colors.
+    Never,
+    /// Cargo will decide whether to output colors based on the tty type.
+    Auto,
+}
+
 /// Builds rustdoc JSON. There are many build options. Refer to the docs to
 /// learn about them all. See [top-level docs](crate) for an example on how to use this builder.
 #[derive(Clone, Debug)]
@@ -224,6 +268,7 @@ pub struct Builder {
     target: Option<String>,
     quiet: bool,
     silent: bool,
+    color: Color,
     no_default_features: bool,
     all_features: bool,
     features: Vec<String>,
@@ -242,6 +287,7 @@ impl Default for Builder {
             target: None,
             quiet: false,
             silent: false,
+            color: Color::Auto,
             no_default_features: false,
             all_features: false,
             features: vec![],
@@ -316,6 +362,13 @@ impl Builder {
         self
     }
 
+    /// Color configuration for the output of `cargo rustdoc`.
+    #[must_use]
+    pub const fn color(mut self, color: Color) -> Self {
+        self.color = color;
+        self
+    }
+
     /// Whether or not to pass `--target` to `cargo rustdoc`. Default: `None`
     #[must_use]
     pub fn target(mut self, target: String) -> Self {
@@ -375,8 +428,12 @@ impl Builder {
         self
     }
 
-    /// Generate rustdoc JSON for a library crate. Returns the path to the freshly
+    /// Generate rustdoc JSON for a crate. Returns the path to the freshly
     /// built rustdoc JSON file.
+    ///
+    /// This method will print the stdout and stderr of the `cargo rustdoc` command to the stdout
+    /// and stderr of the calling process. If you want to capture the output, use
+    /// [`Builder::build_with_captured_output()`].
     ///
     /// See [top-level docs](crate) for an example on how to use it.
     ///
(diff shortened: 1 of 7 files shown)


rustdoc-json/src/builder.rs
```diff
@@ -214,16 +228,23 @@ fn library_name(
...
-pub struct Builder {
+#[derive(Debug)]
+pub struct Builder<'a> {
     toolchain: Option<String>,
     manifest_path: PathBuf,
     target_dir: Option<PathBuf>,
     target: Option<String>,
     quiet: bool,
     silent: bool,
+    capture_output: Option<CaptureOutput<'a>>,
```
Emilgardis: instead of an option on `Builder` this could be a closure which would gives access. e.g introduce a new function `run_cargo_rustdoc_with_output(Builder, fn(&mut std::process::Output))`

Enselic:
Hi orium. I'm not against doing something like this if it is needed, but I'd like to understand more about why it's needed. Can you explain in more detail in what way you plan to use this new API please?

orium:
> Can you explain in more detail in what way you plan to use this new API please?

Yes: basically I want to rustdoc to not output anything at all unless the rustdoc build failed.  In that case I want to present a error of my own and provide the rustdoc stderr.

rustdoc-json/src/builder.rs
Enselic: Another option could be to add a new method `Builder::build_and_capture_output()` that returns something like

```
struct BuildResult {
    rustdoc_json_path: PathBuf,
    stdout: Vec<u8>,
    stderr: Vec<u8>,
}
```

In any case I would prefer to avoid adding lifetimes to `struct Builder`. And I don't think it would make sense for `struct Builder` to own output data. We're probably going to need to experiment a bit to find an API that feels right and works well.

Enselic (changes requested):
That makes sense, thanks. I'm not a big fan of adding lifetimes to `struct Builder` so I added a third proposal next to Emilgardis proposal. Hard to tell which one ends up best without experimenting a bit.

rustdoc-json/src/builder.rs
orium: I did consider something like Enselic suggested. I'll implement that.

edit: Oh, we can't do that: we need the stdout/stderr even if we return `Err(_)`. It is when things failed it is more likely the user wants to look at the stderr. So we can do something like Emilgardis suggested, but I think it is better to pass two `&mut Vec<u8>` instead of a closure. This is because the caller will likely want to know the result before action anything with the output, meaning the closure will have to save the state and to wait until the `build_capture_output()` returns to inspect if succeeded or failed.

orium:
I've implemented like I discussed in https://github.com/Enselic/cargo-public-api/pull/574#discussion_r1599059467. Didn't introduce a new pub function for "build and capture", simply just added `Builder::build_capture_output()`.

rustdoc-json/src/builder.rs
Emilgardis: I'm not sure I follow what you mean with

> This is because the caller will likely want to know the result before action anything with the output, meaning the closure will have to save the state and to wait until the build_capture_output() returns to inspect if succeeded or failed.

isn't that true for the current design? My thinking with using a closure is that it would look like

```rs
let mut stderr = Vec::new();
let maybe_path = run_cargo_rustdoc_with_output(builder, |output| stderr.extend(output.stderr));
```

orium: > isn't that true for the current design?

Yes, it is. What I'm saying is that the extra versatility of a closure will, in general, not be very useful because I think most use cases will be exactly like your example: store the output in a vector and decide what to do with it after checking the success or failure of `run_cargo_rustdoc_with_output()`. In which case it is a bit more ergonomic to simply pass `&mut Vec<u8>` to `run_cargo_rustdoc_with_output()`.

I don't feel strongly about this though: I think both options are reasonable.

rustdoc-json/src/builder.rs
```diff
@@ -387,6 +409,18 @@ impl Builder {
     pub fn build(self) -> Result<PathBuf, BuildError> {
         run_cargo_rustdoc(self)
     }
+
+    /// Generate rustdoc JSON for a library crate. This works like [`Builder::build()`], but will
+    /// capture the stdout and stderr of the `cargo rustdoc` command. The output will be stored in
+    /// the parameters `stdout` and `stderr`.
+    pub fn build_capture_output(
+        self,
+        stdout: &mut Vec<u8>,
```
Enselic: Since `Vec<u8>` implements `Write`, wouldn't it be a good idea to change these to `impl Write`? And then maybe add a rustdoc example that uses `Vec<u8>` to make it clear that is possible.

Also, can you run ./scripts/bless-expected-outout-for-tests.sh please? So the PR includes the diff of the public API of the rustdoc-json crate.

orium: yeah, using `Write` is neat. Done.

rustdoc-json/src/builder.rs
```diff
@@ -387,9 +420,33 @@ impl Builder {
     pub fn build(self) -> Result<PathBuf, BuildError> {
         run_cargo_rustdoc(self)
     }
+
+    /// Generate rustdoc JSON for a library crate. This works like [`Builder::build()`], but will
+    /// capture the stdout and stderr of the `cargo rustdoc` command. The output will be written to
```
Enselic, line 451: Let's clarify somewhere in the docs something like:

> Note: It is the stdout and stderr output of the command that builds rustdoc JSON that is captured, not the rustdoc JSON itself.

I think you already explain it, but I think it makes sense to be extra clear and put more emphasis on it since it seems like something that can be a bit confusing.

rustdoc-json/src/builder.rs
```diff
@@ -387,9 +420,33 @@ impl Builder {
...
+    /// #
+    /// let mut stderr: Vec<u8> = Vec::new();
+    ///
+    /// let json_path: PathBuf = rustdoc_json::Builder::default()
+    ///     .toolchain("nightly")
+    ///     .manifest_path("Cargo.toml")
+    ///     .build_capture_output(std::io::sink(), &mut stderr)
+    ///     .unwrap();
+    /// ```
+    pub fn build_capture_output(
```
Enselic: I'd like to call this `build_with_captured_output(...)` which I think is a bit clearer.

rustdoc-json/src/builder.rs
```diff
@@ -33,11 +60,17 @@ pub fn run_cargo_rustdoc(options: Builder) -> Result<PathBuf, BuildError> {
         if manifest.package.is_none() && manifest.workspace.is_some() {
             Err(BuildError::VirtualManifest(options.manifest_path))
         } else {
-            Err(BuildError::General(String::from("See above")))
+            Err(BuildError::CrateBuildError)
```
Enselic: Let's give a hint for future explorers of the code base:

```suggestion
            // In this case the user will see the exact error on stderr
            Err(BuildError::CrateBuildError)
```

Edit: Actually, let's move this comment to the docs of `BuildError::CrateBuildError`

rustdoc-json/src/builder.rs
```diff
@@ -387,9 +420,33 @@ impl Builder {
     pub fn build(self) -> Result<PathBuf, BuildError> {
```
Enselic, line 446: Add a note saying something like "If you want to capture stdout and stderr from the command that builds rustdoc JSON you can use [`build_with_captured_output`]"

Enselic (changes requested):
I like this API, thanks! I added some small comments. Now the only thing missing for merge as far as I'm concerned is some form of regression test to make sure this feature keeps working and does not accidentally break.

I'm puzzled by your CI failure. A no-op PR (https://github.com/Enselic/cargo-public-api/pull/575) did not fail, but the failure you see seems unrelated to your PR. Maybe a temporary fluke that will be fixed by rebase?

rustdoc-json/src/builder.rs
```diff
@@ -387,9 +420,33 @@ impl Builder {
...
+    ///
+    /// ```no_run
+    /// # use std::path::PathBuf;
+    /// #
+    /// let mut stderr: Vec<u8> = Vec::new();
+    ///
+    /// let json_path: PathBuf = rustdoc_json::Builder::default()
+    ///     .toolchain("nightly")
+    ///     .manifest_path("Cargo.toml")
+    ///     .build_capture_output(std::io::sink(), &mut stderr)
```
Enselic: Nice example btw!

rustdoc-json/src/lib.rs
```diff
@@ -38,6 +38,10 @@ pub enum BuildError {
     #[error("Failed to build rustdoc JSON: {0}")]
     General(String),
 
+    /// An error originating from building the crate's documentation.
```
Enselic: Let's move my comment here:

```suggestion
    /// An error originating from building the crate's documentation.
    /// In this case the exact error can be seen on stderr.
```

rustdoc-json/src/lib.rs
```diff
@@ -38,6 +38,10 @@ pub enum BuildError {
...
+    #[error("Failed to build crate")]
```
Enselic:
```suggestion
    #[error("Failed to build crate (see stderr)")]
```

rustdoc-json/src/builder.rs
orium: Explained it better with a more specific example.

rustdoc-json/src/builder.rs
orium: Done.

rustdoc-json/src/builder.rs
orium: Done.

rustdoc-json/src/lib.rs
orium: Done

rustdoc-json/src/lib.rs
orium: That is not the case when the the output is captured. Left that information in the error variant's description. Let me know if you want the message itself to be more verbose.

rustdoc-json/src/builder.rs
```diff
@@ -33,7 +60,7 @@ pub fn run_cargo_rustdoc(options: Builder) -> Result<PathBuf, BuildError> {
         if manifest.package.is_none() && manifest.workspace.is_some() {
             Err(BuildError::VirtualManifest(options.manifest_path))
         } else {
-            Err(BuildError::General(String::from("See above")))
+            Err(BuildError::CrateBuildError)
```
Enselic: (Obsolete accidental comment, please ignore.)

rustdoc-json/src/lib.rs
Enselic: The use case I have in mind is this:

Someone completely new to the library gets the "Failed to build crate" error. That is very frustrating, because a good error will say *why* there was an error. I don't think that someone should have to go look at the rustdoc for the enum. So I think we at least need a hint on where to look next.

The phrase "see stderr" is in my mind short for "see stderr, regardless of it is printed in your terminal or if you captured it yourself". The way I see it, stderr needs to be looked at in both cases. If  you strongly disagree and want to add a more verbose description in the error in that case, that's OK with me.

But in one way or another I think we should refer to stderr in the error message.

Enselic (changes requested):
I did a more detailed review and inlined a function and made the CI failure pass. But it looks like your test fails too.

Feel free to squash my commits into your commit without attribution to me. Just seemed easier to push the commits myself to show what I mean rather than writing comments about it.

I think I might have some more comments later on the details, but now I think the next step is to make CI green. I could reproduce the failure locally.

rustdoc-json/src/builder.rs
Enselic: Oops, I changed my mind and meant to remove this comment, but forgot about it. Please ignore this comment for now.

orium:
> The phrase "see stderr" is in my mind short for "see stderr, regardless of it is printed in your terminal or if you captured it yourself".

Ok, I wasn't reading in that way, but that makes sense.

> Feel free to squash my commits into your commit without attribution to me.

Squashed.

> I did a more detailed review and inlined a function and made the CI failure pass. But it looks like your test fails too.

Thanks for fixing `custom_toolchain_via_proxy`. My test was failing because the github's CI tty somehow makes cargo color the output, which doesn't happen when I run the test in my machine. I could just strip colors from the output, but actually this is something we want users to be able to control, so I've added `Builder::color()` to control that.

Enselic (commented):
(I've been unusually busy in recent days, hence the slow reply.)

I added some more error enums (feel free to squash with your commit without attribution. It would also be good to rebase on dependabot commits), since otherwise users could get errors like "file not writable" without context, but now they get something like "could not write stderr: file not writable".

I tried to get used to the terminology of "CrateBuildError" for failure to build rustdoc JSON, but I think that has too strong association to "cargo build", which is not what rustdoc JSON is using. So I renamed it to "BuildRustdocJsonError". Is that OK with you?

orium:
> (I've been unusually busy in recent days, hence the slow reply.)

No problem at all!

> I added some more error enums (feel free to squash with your commit without attribution.

Given that your commit contains non-trivial changes I think it makes sense to preserve it.

> It would also be good to rebase on dependabot commits)

Rebased.

> I tried to get used to the terminology of "CrateBuildError" for failure to build rustdoc JSON, but I think that has too strong association to "cargo build", which is not what rustdoc JSON is using. So I renamed it to "BuildRustdocJsonError". Is that OK with you?

Yeah, happy with that.

Enselic (approved):
Thanks, let's merge! I hope to make a release relatively soon.

orium:
Thank you! Looking forward for the release. Cheers!

Enselic:
Released as [v0.9.1](https://crates.io/crates/rustdoc-json/0.9.1) 🚀 

(You will be credited by name next time I do a cargo-public-api [release](https://github.com/Enselic/cargo-public-api/releases), but that might take a while)

orium:
That's great. Thank you!

PR #575, opened 2024-05-17
CI test: Does CI fail with no-op change?

Enselic:
https://github.com/Enselic/cargo-public-api/pull/574 fails in strange way. Does this PR too?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# cargo-public-api
+## cargo-public-api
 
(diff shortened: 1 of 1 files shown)


Enselic:
The problem was https://github.com/Enselic/cargo-public-api/pull/574/commits/e33c17abbb40a6636a99e8776e77852d45f9d8cb

PR #580, opened 2024-05-24
 Match `cargo`'s handling of spaces with `--features`

### LINKED ISSUE #573: Features in combination with diff
I'm trying to use public-api with features in the diff mode. However, `--features` eats all following strings, e.g.:
```
cargo public-api --features runtime-tokio-hyper,async diff HEAD..master 
```
This runs the normal public-api with features `runtime-tokio-hyper`, `async` `diff` `HEAD..master`.

Is there any way to provide these features? I tried quoting and with or without `,` between.

Enselic:
Closes #573

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -57,7 +57,7 @@ pub struct Args {
     omit: Option<Vec<Omit>>,
 
     /// Space or comma separated list of features to activate
-    #[arg(global = true, long, short = 'F', num_args = 1..)]
+    #[arg(global = true, long, short = 'F')]
     features: Vec<String>,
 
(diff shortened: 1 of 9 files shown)


Emilgardis (approved):
lgtm!


a side note, the two empty diff outputs are a bit useless imo, maybe we could symlink them?

Enselic:
> a side note, the two empty diff outputs are a bit useless imo, maybe we could symlink them?

Good point, no reason to duplicate them. I replaced them with `cargo-public-api/tests/expected-output/no_diff.txt` which can be used for both tests (and other tests in the future). (I think it's nice to avoid symlinks when possible.)

PR #581, opened 2024-06-07
Support `nightly-2024-06-07`

Enselic:
https://github.com/rust-lang/rust/pull/125958 changed rustdoc JSON format, and this PR upgrades stuff to support the new format.


Enselic (commented):
(I will merge this right away and make a release to minimize trouble for users)

PR #589, opened 2024-06-27
Parallelize tests by introducing internal testsuite crate

Enselic:
This PR is inspired by https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html.

Parallelize tests by introducing internal testsuite crate and move most of our tests there. 

# Before vs after the PR

I checked with
```sh
time cargo test | grep '^test.*\.\.\.' | wc -l
```
and before this PR we run 152 tests. After, we also run 152 tests. So nothing seems lost.

In the upside, after this PR, repated `cargo test` runs takes 6 seconds on my machine. Before the PR it took 12 seconds.


Enselic:
I changed my mind and will abandon this PR. Mostly because we can't move cargo-public-api tests to testsuite if we don't add a build-script that makes sure the binary is built before tests run. And in retrospect the approach in this PR feels a bit contrived and special-casey. Maybe some other day I will take another shot at something similar, but for now I am abandoning this.

PR #609, opened 2024-08-18
Bless `nightly-2024-08-18` output

EnselicCICD:
Automatically created by https://github.com/cargo-public-api/cargo-public-api/actions/runs/10437554908/jobs/auto-bless


Enselic (approved):
https://github.com/rust-lang/rust/pull/126877

PR #613, opened 2024-09-06
Add an option to disable toolchain checks

Vaiz:
We aim to integrate this tool into our company’s CI pipeline for automated checks. However, we face an issue due to our custom `rustup`, which has a different name and command set compared to the official one. All of that doesn’t give a chance for `cargo-public-api` to run.

Fortunately, the toolset supports nightly features and functions correctly when running cargo directly without specifying a toolset.

This PR introduces an environment variable, CARGO_PUBLIC_API_IGNORE_TOOLCHAIN, to disable toolset checks, allowing cargo-public-api to run seamlessly with our custom setup.

### DIFF
--- a/cargo-public-api/src/main.rs
+++ b/cargo-public-api/src/main.rs
@@ -573,7 +573,10 @@ fn subcommand_name(bin: OsString) -> Option<OsString> {
 fn resolve_toolchain(args: Args) -> ArgsAndToolchain {
     // This option allows the use of a custom version of rustup that is pre-configured to run nightly
     if std::env::var_os("CARGO_PUBLIC_API_IGNORE_TOOLCHAIN").is_some() {
-        return ArgsAndToolchain { args, toolchain: None };
+        return ArgsAndToolchain {
+            args,
+            toolchain: None,
+        };
     }
 
     let toolchain = if toolchain::is_probably_stable() {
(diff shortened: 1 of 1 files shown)


Emilgardis:
What exactly is it failing on? Do your cargo/rustc/rustdoc binaries not support `+toolchain` maybe? I think we can solve this in another way thats also applicable to a normal rustup setup

Vaiz:
Because we have a different name instead of rustup, it fails to find it
`Error: Failed to build rustdoc JSON: required program rustup not found in PATH. Is it installed?`

Another problem is that our rustup doesn't support `run` command.

Vaiz:
I've considered other options, but none seemed good enough. If you have an alternative in mind, I would be glad to try it

Emilgardis:
Ok! How about if the `RUSTDOC`/`CARGO` variable is set, always respect it. We'll have to be careful with this however due to build-scripts setting this envvar, but you shouldnt really use our cargo binary in a buildscript

Vaiz:
> Ok! How about if the `RUSTDOC`/`CARGO` variable is set, always respect it. We'll have to be careful with this however due to build-scripts setting this envvar, but you shouldnt really use our cargo binary in a buildscript

That should work, although changes won't be so subtle

Vaiz:
After further examining the code, I found a workaround. It’s possible to call `rustdoc` manually and then feed the output to `public-api` via the `rustdoc_json` argument.

I can use this workaround for now, but it would be nice to use public-api directly in the future.

Vaiz:
Hi @Emilgardis ,
Thank you for your feedback and suggestions on this PR. I agree that it might be better to handle this differently, I propose that we close this PR without merging and open a tracking issue to support the `RUSTDOC`/`CARGO` variables.

Emilgardis:
I've opened #624

PR #617, opened 2024-09-13
Support `nightly-2024-09-10` and later

### LINKED ISSUE #616: Failure with `nightly-2024-09-10` and later due to renamed rustdoc JSON elements
With recent rust nightly build (nightly-2024-09-10), public-api commands are failing on upstream projects (https://github.com/aya-rs/aya and https://github.com/bpfman/bpfman are hitting issue). Looks like rust-lang, via [PR130133](https://github.com/rust-lang/rust/pull/130133) pulled in [PR128667](https://github.com/rust-lang/rust/pull/128667). I believe https://github.com/aDotInTheVoid/rustdoc-types needs to be updated to take into account these changes and a new release created (see [rustdoc-types:Issue32](https://github.com/aDotInTheVoid/rustdoc-types/issues/32)) and then this repo needs to pull in that rustdoc-types release.

Billy99:
rust-lang changed names of several of it's field namess which is causing issues in projects using cargo-public-api. Bumping rustdoc-types to the latest release, v0.30.0.

Fixes: #616


public-api/src/item_processor.rs
```diff
@@ -357,9 +357,9 @@ pub(crate) fn sorting_prefix(item: &Item) -> u8 {
             ImplKind::Blanket => 24,
         },
 
-        ItemEnum::ForeignType => 25,
+        ItemEnum::ExternType => 25,
 
-        ItemEnum::OpaqueTy(_) => 26,
+        // ItemEnum::OpaqueTy(_) => 26, OpaqueTy removed from rustdoc-types
```
Billy99: `OpaqueTy` was deleted from rustdoc-types. Does `TraitAlias` move to 26 or is numbering preserved? If numbering is preserved, should this just be commented out to indicated it's been deleted or just remove the line?

public-api/src/render.rs
```diff
@@ -940,6 +939,7 @@ impl<'c> RenderingContext<'c> {
                 output
             }
             GenericBound::Outlives(id) => vec![Token::lifetime(id)],
+            GenericBound::Use(_) => todo!(),
```
Billy99: Need some help on what needs to be done here. `GenericBound::Use(Vec<String>)` was added in v0.28.0 of rustdoc-types.

Billy99 (commented):
Trying to help out with some grunt work, but had two questions.

public-api/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub use public_item::PublicItem;
 /// The rustdoc JSON format is still changing, so every now and then we update
 /// this library to support the latest format. If you use this version of
 /// nightly or later, you should be fine.
-pub const MINIMUM_NIGHTLY_RUST_VERSION: &str = "nightly-2024-07-05";
+pub const MINIMUM_NIGHTLY_RUST_VERSION: &str = "nightly-2024-09-10";
```
Billy99, line 65: I "think" this needs to be updated? It passed CI without changing, but the rustdoc JSON format changed so guessing this needs to be bumped.

README.md
```diff
@@ -142,7 +142,8 @@ cargo public-api -sss
 
 | Version          | Understands the rustdoc JSON output of  |
 | ---------------- | --------------------------------------- |
-| 0.37.x —         | nightly-2024-07-05 —                    |
+| 0.38.x —         | nightly-2024-09-10 —                    |
+| 0.37.x — 0.37.x  | nightly-2024-07-05 — nightly-2024-09-09 |
```
Enselic: Let's simplify to this:

```suggestion
| 0.37.x           | nightly-2024-07-05 — nightly-2024-09-09 |
```

cargo-public-api/MINIMUM_NIGHTLY_RUST_VERSION_FOR_TESTS
```diff
@@ -1 +1 @@
-nightly-2024-08-18
+nightly-2024-09-10
```
Enselic: This file can be removed now. It is only needed when the version for tests is different from the minimum version for clients. (See https://github.com/cargo-public-api/cargo-public-api/blob/main/docs/RELEASE.md#bumping-minimum_nightly_rust_version.)

public-api/src/item_processor.rs
Enselic: We have a good set of regression tests. If CI passes then you (and me) can be confident that nothing is broken.

My preference would be to remove the line but keep other numbers, because:
1. git blame becomes less noisy
2. it can be useful to have holes in the numbering in the future if new stuff needs to be added in the middle

public-api/src/lib.rs
Enselic: Yep this should be updated. Strange that CI passed without it. I would expect tests to fail since now rustdoc-types should fail to parse the old JSON that the old nightly generates.

public-api/src/render.rs
Enselic: As long as CI is green we can leave it as it is. I'm too lazy (read: short of time) to figure out the syntax to trigger this code path. It might require a nightly-only rustc feature, in which case it is out of scope of this project.

If this can be triggered with stable Rust code, someone will notice and file an issue, and then we'll know how to trigger this code path, and can add a regression test.

Let's change the todo!() to be more descriptive though. Something like:

```rs
todo!("Can this code path be triggered with just stable Rust? If you hit this todo with just stable Rust, please file an issue with a minimal reproducer. Thanks!");
```

Enselic (changes requested):
Thanks! Just had some small comments.

Once this is merged it would be great if you could go through our "how to release"-document found at https://github.com/cargo-public-api/cargo-public-api/blob/main/docs/RELEASE.md and then make a separate PR with all the changes. We'll need one PR per sub-project, in case we need to make releases to more projects than public-api.

Once each PR is merged, a maintainer is just a couple of button clicks away from making a release.

But let's get this PR merged before we take that step. (If you are interested. It would be greatly appreciated.)

public-api/src/item_processor.rs
Billy99: Removed.

public-api/src/render.rs
Billy99: Done.

README.md
Billy99: Done.

cargo-public-api/MINIMUM_NIGHTLY_RUST_VERSION_FOR_TESTS
Billy99: File removed.

Enselic (approved):
👍 let's merge

PR #622, opened 2024-09-16
RELEASE.md: Clarify when not to release

Enselic:
CC @Billy99 

Also see https://github.com/cargo-public-api/cargo-public-api/pull/619 and https://github.com/cargo-public-api/cargo-public-api/pull/621.

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -21,7 +21,7 @@ If `MINIMUM_NIGHTLY_RUST_VERSION` must be bumped then by necessity both `cargo-p
 
 ### When not to release
 
-* If a package (e.g. `rustup-toolchain`, `rustdoc-json`, etc) has had no semver breaking updates to any regular `dependencies` (i.e. we disregard `dev-dependencies`) since the last release, then we don't need to make a new release of that package. Because there are no techinical reasons to do it. On the contrary, it creates unnecessary churn for downstream users.
+* If a package (e.g. `rustup-toolchain`, `rustdoc-json`, etc) has had no semver breaking updates to any regular `dependencies` (i.e. we disregard `dev-dependencies` changes in Cargo.toml files) since the last release, then we don't need to make a new release of that package. Because there are no technical reasons to do it. On the contrary, it creates unnecessary churn for downstream users.
 
 ## How to release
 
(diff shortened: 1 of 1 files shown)


docs/RELEASE.md
```diff
@@ -20,6 +20,10 @@ If `MINIMUM_NIGHTLY_RUST_VERSION` must be bumped then
   rm cargo-public-api/MINIMUM_NIGHTLY_RUST_VERSION_FOR_TESTS
   ```
 
+### When not to release
+
+* If a package (e.g. `rustup-toolchain`, `rustdoc-json`, etc) has had no semver breaking updates to any regular `dependencies` (i.e. we disregard `dev-dependencies`) since the last release, then we don't need to make a new release of that package. Because there are no techinical reasons to do it. On the contrary, it creates unnecessary churn for downstream users.
```
Billy99:
```suggestion
* If a package (e.g. `rustup-toolchain`, `rustdoc-json`, etc) has had no semver breaking updates to any regular `dependencies` (i.e. we disregard `dev-dependencies` changes in Cargo.toml files) since the last release, then we don't need to make a new release of that package. Because there are no technical reasons to do it. On the contrary, it creates unnecessary churn for downstream users.
```

Billy99 (commented):
Just a nit

docs/RELEASE.md
```diff
@@ -20,6 +20,10 @@ If `MINIMUM_NIGHTLY_RUST_VERSION` must be bumped then
...
+
 ## How to release
 
 ### `cargo-public-api`
```
Billy99, line 28: While your are making changes to this file, in this section (`cargo-public-api`), the 4th bullet reads:
```
1. For each PR included in the release, adjust its `[category-*]` according to [release.yml](https://github.com/cargo-public-api/cargo-public-api/blob/main/.github/release.yml) and tweak the PR title if necessary to turn it into good release note entry.
```
Suggestion to change to:
```
1. For each PR included in the release, adjust its `[category-*]` according to [release.yml](https://github.com/cargo-public-api/cargo-public-api/blob/main/.github/release.yml) by adding or changing the label associated with the PR. Also tweak the PR title if necessary to turn it into good release note entry.
```

Note: I am not able to change the labels on a PR, so this may need to updated to indicate that only Maintainers can do this step.

Enselic: Marking as resolved since I removed this text in https://github.com/cargo-public-api/cargo-public-api/pull/644

PR #623, opened 2024-09-16
Bump public-api to v0.38.0

Billy99:
Changes to rustdoc-types required updates to repo.

### DIFF
--- a/public-api/CHANGELOG.md
+++ b/public-api/CHANGELOG.md
@@ -1,5 +1,9 @@
 # `public-api` changelog
 
+## v0.38.0
+* Support `nightly-2024-09-10`, which brought in lots of field renaming due to https://github.com/rust-lang/rust/pull/128667.
+* Removed `OpaqueTy` since it was removed from rustdoc-types.
+
 ## v0.37.0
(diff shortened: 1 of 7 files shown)


public-api/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # `public-api` changelog
 
+## v0.38.0
+* Support `nightly-2024-09-10`, which brought in lots of field renaming due to https://github.com/rust-lang/rust/pull/128667.
+* Removed `OpaqueTy` since it was removed from rustdoc-types.
```
Enselic: nit: I don't think this is noticeable by users, so I think it would be better to remove this line from the changelog

(I will do that and then merge)

Enselic: I ended up simplifying it even futher.

Enselic (approved):
Thanks! Let's merge.

Let's also prepare cargo-public-api for release so we can release both cargo-public-api and public-api at the same time, to make the compatibility matrix consistent (sorry for contradicting myself from before. I will further update release instructions )

PR #625, opened 2024-09-17
Bump cargo-public-api to v0.38.0

Billy99:
Changes to rustdoc-types required updates to repo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,7 +178,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-public-api"
-version = "0.37.0"
+version = "0.38.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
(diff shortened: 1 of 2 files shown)


Billy99:
@Enselic Sorry to bug you but trying to move this along so I can unblock Aya CI. Is there anything else you need me to do? There are two PRs I don't have permission to set the label for (#609 and #623). But I think everything else is ready?

Enselic:
No worries, I understand.

I've got some other urgent private matters to take care of, but I'll look at this as aoon as I get time.

To unblock Aya CI you should be able to use a fixed version of nightly rather than the latest. That is also a general recommendation since then sudden failures like these when the format changes will be not bring down your CI.

To my co-maintainers if they happen to read this: feel free to make a release if you want, after reviewing pending PRs, etc.

Billy99:
Thanks for the update.
FYI: I did pin nightly on one project, bpfman. I had issues pinning on Aya. I got past public-api, but had issues with another tool, Miri, which I'm not very familiar with. Instead of spending a day debugging that, I thought I was just help get this release out. But I'll spend some time on Miri, for when something like this happens again.

Enselic (approved):

PR #634, opened 2024-10-15
Debug why 'Preview release notes' do not auto-trigger

Enselic:
Something is broken here...


Enselic:
I think I will replace our current release notes with https://crates.io/crates/parse-changelog and ask each contributor to add themselves. Then the release notes will be taken directly from the changelog, which will be up to date.

PR #648, opened 2024-11-14
Handle `TraitBoundModifier`s

smoelius:
Hello. I think `TraitBoundModifier`s were not being handled correctly. The present PR is an attempt to fix that. Nits are welcome.


smoelius:
The failing tests pass for me locally. Any idea why they would fail in CI?

Enselic:
Are you using latest nightly locally?

smoelius:
An error on my part meant not all of the test files were updated. The problem should be fixed now.

cargo-public-api/tests/expected-output/omit-auto-derived-impls.txt
```diff
@@ -16,11 +16,11 @@ pub fn example_api::Struct::try_from(value: U) -> core::result::Result<T, <T as
 impl<T, U> core::convert::TryInto<U> for example_api::Struct where U: core::convert::TryFrom<T>
 pub type example_api::Struct::Error = <U as core::convert::TryFrom<T>>::Error
 pub fn example_api::Struct::try_into(self) -> core::result::Result<U, <U as core::convert::TryFrom<T>>::Error>
-impl<T> core::any::Any for example_api::Struct where T: 'static + core::marker::Sized
+impl<T> core::any::Any for example_api::Struct where T: 'static + ?core::marker::Sized
```
Enselic, line 19: rustdoc indeed renders a ? in this case, so we should too 👍 

image

public-api/src/render.rs
```diff
@@ -937,10 +937,20 @@ impl<'c> RenderingContext<'c> {
...
+                modifier,
             } => {
                 let mut output = vec![];
                 output.extend(self.render_higher_rank_trait_bounds(generic_params));
+                match modifier {
+                    TraitBoundModifier::None => {}
+                    TraitBoundModifier::Maybe => {
+                        output.push(Token::symbol("?"));
+                    }
+                    TraitBoundModifier::MaybeConst => {
```
Enselic: I think this is a nightly only feature? If so it is out of scope of this tool, since it would result in too much churn for us to support nightly features. So I think I would prefer to treat this the same way as `TraitBoundModifier::None`, but I am open to being convinced otherwise.

public-api/src/render.rs
```diff
@@ -1354,6 +1364,25 @@ mod test {
         );
     }
 
+    #[test]
+    fn test_maybe_sized() {
```
Enselic: Thanks for adding an additional test! I do however think this is too much of a "test how things are done" rather than a "test what the end result is", so I think I would prefer to remove it. The case is well-tested in our existing test suite, so maintaining this "implementation-detail test" is likely to be unnecessary overhead going forward. Sorry for the negativity, but I value long-term maintainability of this project high. I should probably remove the existing similar tests we have...

Enselic (changes requested):

Enselic:
> An error on my part meant not all of the test files were updated. The problem should be fixed now.

Just a tip in case you didn't see it yet: Typically you only need to run `scripts/bless-expected-output-for-tests.sh` to update expected output for tests.

Speaking of scripts, `scripts/run-ci-locally.sh` is also good to know about.

public-api/src/render.rs
```diff
@@ -937,10 +937,20 @@ impl<'c> RenderingContext<'c> {
...
                 generic_params,
-                ..
+                modifier,
             } => {
                 let mut output = vec![];
                 output.extend(self.render_higher_rank_trait_bounds(generic_params));
+                match modifier {
+                    TraitBoundModifier::None => {}
+                    TraitBoundModifier::Maybe => {
+                        output.push(Token::symbol("?"));
```
Enselic, line 947: Also: Can you add an entry about this to CHANGELOG.md and public-api/CHANGELOG.md please? The same entry can be used in both files.

Then I can make a release immediately after this PR is merged.

Enselic: Here is my proposal on the entry:

```
## v0.41.0
* Render `?` in front of `core::marker::Sized` if applicable.
```

public-api/src/render.rs
Enselic: Other tests removed in https://github.com/cargo-public-api/cargo-public-api/pull/652.

smoelius:
I think I made all of the requested changed. Please let me know if I missed anything, or if you see something else.

Enselic (approved):
Looks good, thank you.

Enselic:
Released as https://github.com/cargo-public-api/cargo-public-api/releases/tag/v0.41.0

PR #649, opened 2024-11-14
Eliminate unnecessary `.map(Deref::deref)`

smoelius:
A small change for the sake of concision.


Enselic (approved):
Code simplifications are always nice, thanks

PR #654, opened 2024-11-15
Add trademark notice to READMEs with Rust or Cargo in the project name

Enselic:
To fulfil requirements outlined in the Rust Project's trademark policy, more specifically "Updated Rust Trademark Policy Draft - November 2024". See https://drive.google.com/file/d/1hjTx11Fb-4W7RQLmp3R8BLDACc7zxIpG/view.

(I got help from ChatGPT to formulate the notice.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,3 +164,7 @@ See [CONTRIBUTING.md](./docs/CONTRIBUTING.md).
 - [Enselic](https://github.com/Enselic)
 - [douweschulte](https://github.com/douweschulte)
 - [Emilgardis](https://github.com/Emilgardis)
+
+# Trademark Notice
+
+"Rust" and "Cargo" are trademarks of the Rust Foundation. This project is not affiliated with, endorsed by, or otherwise associated with the Rust Project or Rust Foundation.
--- a/rustdoc-json/README.md
+++ b/rustdoc-json/README.md
@@ -26,3 +26,7 @@ Please refer to [CHANGELOG.md](https://github.com/cargo-public-api/cargo-public-
 ## Tests
 
 This library is indirectly and heavily tested through the [`public-api`](https://crates.io/crates/public-api) and [`cargo-public-api`](https://crates.io/crates/cargo-public-api) test suites. Their tests heavily depend on this library, so if all of their tests pass, then this library works as it should. All tests are of course ensured to pass before a new release is made.
+
+# Trademark Notice
+
+"Rust" and "Cargo" are trademarks of the Rust Foundation. This project is not affiliated with, endorsed by, or otherwise associated with the Rust Project or Rust Foundation.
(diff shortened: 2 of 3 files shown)


Emilgardis:
I dont think this change was needed, but i guess it doesnt hurt

Enselic:
Hmm don't you think? Based on

> The most basic rule is that the Rust trademarks cannot be used in ways that appear (to a casual observer) official, affiliated, or endorsed by the Rust Project or Rust Foundation, unless you have written permission from the Rust Foundation.

I think it can be argued that it is needed. To a casual observer (not familiar with the cargo subcommand system) I think "cargo-public-api"  can appear as an official part of Cargo. 

But I'm open to revert this if needed. But as you say I don't think it hurt much, although noise in the readme is always annoying.

Emilgardis:
`cargo-foobar` is explicitly stated in the document to be allowed as long as it's not implied/indicated that it's an official cargo extension or similar, e.g it's reasonable that a casual observer understands that `cargo-*` is how binaries can become subcommands in cargo. Same goes for having rust/cargo/rustup in the title of code repositories or names of crates is fine if it's just there to signify it's compatible with the language.

Enselic:
I agree up to this part:

>  e.g it's reasonable that a casual observer understands that cargo-* is how binaries can become subcommands in cargo.

For me it is not clear that this can be assumed, based on the draft policy. Can you elaborate on how to draw this conclusion from the draft please?

Here are some additional quotes that I interpret as `cargo-public-api` needing the trademark notice:

> The most basic rule is that the Rust trademarks cannot be used in ways that appear [...] official, affiliated, or endorsed by the Rust Project or Rust Foundation, [...] This is a requirement for all uses, even those that are
listed as not requiring explicit approval.

> There are a variety of uses that do not require explicit approval. However, in all of the cases
outlined below, you must ensure that use of the Rust trademarks does not appear official,
affiliated or endorsed, as explained above.

We can also consider invoking this paragraph and reach out for clarification:

> If you have any doubts about whether your intended use of a Rust Trademark requires permission, please contact us.

Emilgardis:
Might be worth reaching out to lessen confusion and make it more clear 👍🏼 (edit: I filled out the form for feedback :) )

I drew the conclusion from

> The most basic rule is that the Rust trademarks cannot be used in ways that appear (to a casual
observer) official, affiliated, or endorsed by the Rust Project or Rust Foundation. 
> [...] For example the use of terms such as ‘official’, ‘original’
or ‘only’, (e.g. ‘The only Rust conference in the USA’ or ‘The original Rust Async Book’), is likely
to give the appearance that the use of the trademark is officially endorsed by the Rust Project or
Rust Foundation, and therefore would not be acceptable.
> ...
> Uses that do not require explicit approval:
> [...] in all of the cases
outlined below, you must ensure that use of the Rust trademarks does not appear official,
affiliated or endorsed, as explained above.
> ...
> * Names for software in the form of "cargo-foobar", where this is a subcommand for
cargo intended to be executed as "cargo foobar". This is allowed, provided it is not
implied or indicated to be an official cargo extension.

From this I draw that for it to look official it has to claim something, the readme and other descriptions don't claim anything in relation to it being an "official" product, we only use the mark to integrate with cargo.

Enselic:
>  I filled out the form for feedback

👍

Issue #655, opened 2024-11-15
Add `id` and `parent_id` functions, and one other small request


smoelius:
As alluded to in #627, I am using a fork of `public-api` to generate wrappers for standard library functions and types.

To avoid using the fork, I was wondering if you would consider the following changes:
- Add an `id` function to return the `rustdoc_types::Id` from which a `PublicItem` was constructed
- Add a `parent_id` function to return the `rustdoc_types::Id` of the `PublicItem`'s parent, if one exists
- Eliminate the `name != "_"` condition in the following code so that a variable name is always emitted: https://github.com/cargo-public-api/cargo-public-api/blob/0fb7f150a88da4eba85e84f1ad12b98cc7258487/public-api/src/render.rs#L468-L470

If you prefer, I could submit a PR to show you the changes I am imagining.

Thanks regardless of what you decide.

Enselic:
I think your proposal is reasonable enough to make it worthwhile to create a draft PR so we can get a feeling for the details before making a decision.

If it looks ok then and if you agree to help maintain this API surface I think it should be fine. (But other maintainers might have stronger objections.)

smoelius:
> I think your proposal is reasonable enough to make it worthwhile to create a draft PR so we can get a feeling for the details before making a decision.

The draft PR is #656.

smoelius:
#656 was merged.

smoelius:
Thanks again!

PR #656, opened 2024-11-15
Add `id` and `parent_id` methods to `PublicItem`

smoelius:
This is a draft PR to facilitate discussion in #655.


cargo-public-api/tests/expected-output/list_public_items.txt
```diff
@@ -134,6 +134,8 @@ impl core::panic::unwind_safe::UnwindSafe for public_api::PublicApi
 pub struct public_api::PublicItem
 impl public_api::PublicItem
 pub fn public_api::PublicItem::grouping_cmp(&self, other: &Self) -> core::cmp::Ordering
+pub fn public_api::PublicItem::id(&self) -> rustdoc_types::Id
```
Enselic, line 137: In general I am skeptical about having foreign types as part of our public API since that means we are not in full control of our API surface, but
* in this case the whole point of the API is to allow integrating with `rustdoc_types` (I assume)
* if `rustdoc_types` changes we typically need to bump 0.x.0 anyway
so in this case I think it's fine.

public-api/src/diff.rs
```diff
@@ -142,10 +142,14 @@ fn bag_to_path_map<'a>(difference: impl Iterator<Item = (&'a PublicItem, usize)>
 
 #[cfg(test)]
 mod tests {
+    use rustdoc_types::Id;
+
     use crate::tokens::Token;
 
     use super::*;
 
+    static DUMMY_ID: Id = Id(1234);
```
Enselic: I think it works to have
```suggestion
    const DUMMY_ID: Id = Id(1234);
```
?

public-api/src/render.rs
```diff
@@ -446,15 +446,15 @@ impl<'c> RenderingContext<'c> {
         output.extend(self.render_generic_param_defs(&generics.params));
 
         // Regular parameters and return type
-        output.extend(self.render_fn_decl(sig));
+        output.extend(self.render_fn_decl(sig, true));
```
Enselic: Let's add a simple function to `test-apis/comprehensive_api/src/functions.rs` which triggers the rendering code to render a `_` so that
* we have a regression test
* it becomes easier to see how the new code affects rendering in practice

public-api/tests/public-api.txt
```diff
@@ -258,6 +258,8 @@ pub fn public_api::PublicApi::from(t: T) -> T
 pub struct public_api::PublicItem
 impl public_api::PublicItem
 pub fn public_api::PublicItem::grouping_cmp(&self, other: &Self) -> core::cmp::Ordering
+pub fn public_api::PublicItem::id(&self) -> rustdoc_types::Id
```
Enselic, line 261: It would be nice with a simple program that uses this new API. I think putting it in `public-api/examples/` makes the most sense. The closer it is to your actual use case, the better. But try to keep it as simple as possible to make it easy to maintain.

Enselic (changes requested):
On a first look I find this to be a nice and straightforward change with a new simple API. Nice.

I understand this is just a draft but I took the liberty to add some comments anyway.

I'd like to wait at least some days or maybe even a week before we merge though to give other maintainers a chance to take a good look at this, since it is a relatively high-impact change on the long-term maintainability of the project.

cargo-public-api/tests/expected-output/list_public_items.txt
```diff
@@ -134,6 +134,8 @@ impl core::panic::unwind_safe::UnwindSafe for public_api::PublicApi
 pub struct public_api::PublicItem
```
Enselic, line 134: If we go ahead with this I propose this new CHANGELOG.md entry (only needed for public-api/CHANGELOG.md):

```
## v0.42.0
* Add `PublicItem::id(&self) -> rustdoc_types::Id` and `PublicItem::parent_id(&self) -> core::option::Option<rustdoc_types::Id>`
```

public-api/src/diff.rs
smoelius: :man_facepalming:

public-api/tests/public-api.txt
smoelius: The example I added is a slight adaptation of `list_public_api`. Nits are welcome.

smoelius:
I hope the "Eliminate `Id` references" commit is okay. I forked `public-api` before `rustdoc-types` made `Id` to copyable. So, with this commit, I removed several `&Id` that I introduced, and a few others.

EDIT: I could submit a separate PR if you prefer.

Enselic:
The smaller PRs the better, so if it is possible to split this out into smaller PRs, that is a good thing.

I'm also curious about the performance impact of this PR. We currently don't have any CI check for performance. If we did I think it would be in the form of cachegrind which I think is independent of hardware it runs on since it counts instructions. That is partly what the Rustc performance checks use: https://github.com/rust-lang/rustc-perf/tree/master/collector

The best we have AFAIK in comparing performance before and after is manually running https://github.com/cargo-public-api/cargo-public-api-benchmarks/blob/main/benches/json_processing.rs with `cargo bench`. I don't care super much about performance since we are pretty fast regular sized crates, but this creates a slowdown of say 100% I think we need to discuss options.

I'll also unmark this as draft so we can see how CI is doing.

public-api/src/intermediate_public_item.rs
```diff
@@ -34,6 +40,16 @@ impl<'c> IntermediatePublicItem<'c> {
         &self.path
     }
 
+    #[must_use]
+    pub fn parent_id(&self) -> Option<Id> {
+        self.parent_id
+    }
+
+    #[must_use]
+    pub fn id(&self) -> Id {
```
Enselic, line 49: Oh, CI already did run on the draft. Hmm why doesn't this trigger `--deny missing_docs` then 🤔 (https://github.com/Enselic/cargo-public-api/blob/c12acbbfe1c5d8a96ba8b04e8706e25920cfaf10/scripts/cargo-clippy.sh#L16-L17)

smoelius: Maybe because `IntermediatePublicItem` isn't visible outside of the package?

Regardless, should I add doc comments to the methods I added to that struct?

smoelius:
> The smaller PRs the better, so if it is possible to split this out into smaller PRs, that is a good thing.

:+1: I submitted the "Eliminate `Id` references" as PR #659. I'll rebase this PR if that one is merged.

Would you like me to submit "allow underscore argument names" as its own PR as well?

public-api/src/intermediate_public_item.rs
Enselic: Ah right, I see now you already had docs on the methods.

Enselic:
> Would you like me to submit "allow underscore argument names" as its own PR as well?

Yes please that would be great, with a 0.42.0 CHANGELOG entry in both CHANGELOG.md and public-api/CHANGELOG.md and the new regression test in functions.rs

public-api/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # `public-api` changelog
 
+## v0.43.0
```
Enselic: I'll wait with releasing so let's add this changelog entry to the existing v0.42.0 section in preparation for that release (in both CHANGELOG.md files)

public-api/tests/public-api.txt
Enselic: The example looks good to me, thanks

public-api/src/public_item.rs
```diff
@@ -34,9 +41,21 @@ impl PublicItem {
         PublicItem {
             sortable_path: public_item.sortable_path(context),
             tokens: public_item.render_token_stream(context),
+            parent_id: public_item.parent_id(),
+            id: public_item.id(),
         }
     }
 
+    /// The [`Id`] of this item's logical parent (if any)
```
Enselic, line 49: I think we also want these two to be `#[must_use]`?

smoelius:
Please let me know if you would like me to squash anything.

Enselic:
Ideally each commit should be self-contained and pass all CI tests. You can check locally with `scripts/run-ci-locally.sh`.

In this case I think I personally would have made the remaining parts of the PR a single commit.

Note that in GitHub you can press the "force-pushed"  link between the comments which allows easy incremental review even with force push, as long as the base commit remains fixed. Although after a while those links gets pruned away.

I'm btw checking the performance of this new code now.

Enselic (approved):
According to my measurements, the performance impact is negligible.

The more I think about this change the more I feel it is OK. So I'll go ahead and mark as Approved. Feel free to force push squashed commits. I'll mark as Approved again. I'll wait some days before merge though to give other maintainers a chance to chime in.  And also to give myself some more time to think this through and maybe review the code one more time ;)

public-api/examples/list_public_functions.rs
```diff
@@ -0,0 +1,27 @@
...
+        .manifest_path("test-apis/example_api-v0.2.0/Cargo.toml")
+        .build()?;
+
+    let public_api = public_api::Builder::from_rustdoc_json(&json_path).build()?;
+
+    let file = File::open(json_path)?;
+    let krate = serde_json::from_reader::<_, Crate>(file)?;
+
+    for public_item in public_api.items() {
+        if !is_function(&krate, public_item.id()) {
```
Enselic: Would you mind moving this code from here to `public-api/tests/public-api-lib-tests.rs` ? I.e. we remove `public-api/examples/list_public_functions.rs` and add a new test to `public-api/tests/public-api-lib-tests.rs`. That way the code is not only compiled in CI but also run.

It would be good if the test used both `PublicItem::id()` and `PublicItem::parent_id()`.

I'm thinking  having something like:

```rs
    let json = rustdoc_json_for_lib(
        r#"
mod a_mod {
    fn a_function() {
    }
}
    "#,
    );
```

and then find the function like in the example and then make sure that the parent is the `a_mod` mod.

smoelius: Please let me know if I misunderstood.

Enselic: Looks good, thanks!

I'll merge this now. I think I'll make a release later this week.

Enselic (approved):

smoelius:
Thanks for being so accommodating!

PR #660, opened 2024-11-16
Render `_` argument names in function declarations

smoelius:
This PR renders `_` argument names as discussed in #656.


Enselic (approved):
You nailed it

PR #676, opened 2024-12-20
README.md: Don't reference CHANGELOG.md

Enselic:
I like READMEs to be as simple as possible, and see references to CHANGELOG.md as noise. If someone is looking for a changelog it is very easy to find the right one.

And let's be honest, very few people read the changelog. So it does not make sense to take up valuable space in the README for it.

(I now see it as a mistake by me to initially add the references.)


Emilgardis:
This is useful for the rendering on crates.io, have no other concerns with removing it

Enselic:
For someone visiting a crates.io page looking for CHANGELOG.md, I think what they will do is click on the Repository link. And then they will see the right CHANGELOG.md. So I think it'll be fine.

Let's merge this and see how many complaints we get : )

PR #682, opened 2025-01-08
Make `./scripts/run-ci-locally.sh` pass with `rustup default nightly`

Enselic:
This makes it easier to prepare for features such as the 2024 edition which is currently only works with a nightly toolchain unless you activate that feature explicitly.

I considered testing this case in our CI but I think it will make CI much slower without a big gain. I think it's better to just fix any regression after the fact.


Enselic:
Sorry for rushing this in but I have pending PRs that depend on this.

PR #694, opened 2025-01-25
public-api: Support for `nightly-2025-01-25`

Enselic:


Enselic:
Rushing this in since otherwise users will see breakage

PR #699, opened 2025-02-05
Do not render function args named ""  (`nightly-2025-02-05` and later)

Enselic:
With `nightly-2025-02-05` the rustdoc JSON can apparently have the arg names be `""` (see https://github.com/cargo-public-api/cargo-public-api/actions/runs/13149774732). Skip rendering such args to keep our regression tests happy.

Note that the rustdoc JSON format itself has not changed, so we don't need to bump miminum required nightly toolchain version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # `cargo-public-api` changelog
 
+## v0.44.0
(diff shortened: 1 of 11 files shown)


Enselic:
Merging directly since Nightly CI fails, but any desired adjustments to the code can be made as a follow-up PR.

PR #700, opened 2025-02-06
Nighty CI: Create an issue if it fails (unless we created an issue already)

Enselic:
So that if a maintainer do not have time for a quick fix, there at least is an issue about that there is a problem.

Disclaimer: I haven't verified that this works, but it should. We'll see.

### DIFF
--- a/.github/workflows/Nightly.yml
+++ b/.github/workflows/Nightly.yml
@@ -23,6 +23,24 @@ jobs:
   ci:
     uses: ./.github/workflows/CI.yml
 
+  # Create an issue if Nightly CI fails unless there already is such an issue.
+  create-issue-on-failure:
+    needs: ci
+    runs-on: ubuntu-latest
+    if: ${{ needs.ci.result == 'failure' }}
+    permissions:
+      issues: write
+    env:
+      GH_TOKEN: ${{ github.token }}
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/Nightly.yml
```diff
@@ -23,6 +23,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: ${{ needs.ci.result == 'failure' }}
+    permissions:
+      issues: write
+    env:
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - run: |
+          gh issue list \
+            --repo cargo-public-api/cargo-public-api | grep "CI failed with " || \
```
Emilgardis, line 38: This feels a bit risky, does this retrieve all open issues?

Enselic: Yep it does but we only have 3?

Enselic: To elaborate a bit more:
The query is limited to the 30 last issues (default number), but
* since I like to keep the issue inbox clean, I don't think we ever will have that many issues
* even if we do the issue it look for can be expected to be relatively new and to always exist on the first "page"

So in practice I think this will work out. Thanks for taking a look anyway!

PR #702, opened 2025-02-08
Avoid circular workspace dependencies (and rewrite CI)

Enselic:
Right now there are circular dependencies in the workspace, which makes making releases feel like a minefield. Making releases must be effortless and fun, so let's stop having circular dependencies.

Start treating rustup-toolchain and rustdoc-json as separate workspaces. We can kind of think of them as living in separate git repos, but they just happen to actually live in this git repo.

### Some other consequences of this change
 * We now need a VS Code workspace for all Cargo workspaces (cargo-public-api.code-workspace) or equivalent for other IDEs. In particular so that Cargo.lock automatically gets updated before run-ci-locally.sh is run.
 * Rewritten CI
 * Less duplication betweeen run-ci-locally.sh and CI.yml
 * Easier to run isolated CI tests for crates locally
 * We get rid of ubuntu-22.04 because we use taiki-e/install-action now

### Todo
- [ ] Test things more thoroughly
- [x] Try using taiki-e/install-action also for shfmt which takes 1 min to install with apt
- [ ] Make sure that making releases is actually pleasant now
    - [ ] Handle that `( cd rustdoc-json ; cargo set-version -p rustdoc-json x.y.z )` does not update `public-api/Cargo.toml`
- [x] Actually make CI pass in GitHub Actions
- [x] Make sure we run just as many tests as before (by e.g. comparing stdout before and after this change)
- [x] Benchmark CI before and after (might not matter much though) (Edit: Approx 2 min both before and after.)

### Possible later improvements
 * Make a single "release all crates sequentally" GitHub Actions script (which checkboxes for exactly what crates to release)


Emilgardis:
Im not sure i follow what the issue is, where is it circular?

Enselic:
Before this change we have this workspace dependency situation (arrow points to dependencies):

```
/--------- cargo-public-api
|          |              |
|          |              |
|          v              v
|     public-api <-----  rustup-toolchain
|     ^     |                 |
|     |     |                 |
|     |     v                 |
\-->  rustdoc-json   <--------/
```

After this change we have:

```
/--------- cargo-public-api
|          |              |
|          |              |
|          v              v
|     public-api         rustup-toolchain
|           |
|           |
|           v
\-->  rustdoc-json
```

E.g. rustup-toolchain still depends on public-api, but from crates.io, like a "regular" package.

One of the main goals of this change is to allow to make a single "update stuff for a new release" commit, and then trigger a single "create releases in the right order" GitHub Action. So that making a release is just
1. Create one commit (with bumped versions etc)
2. Click a GitHub Actions button

Right now one has to do separate commits and wait for dependencies to be released to avoid circular dependencies. So making a full set of release takes quite some time because you have to baby-sit the release process.

The most recent example of the problem was in https://github.com/cargo-public-api/cargo-public-api/actions/runs/12966432098 where I could not release rustup-toolchain which _should_ be at the bottom of the stack and always be releasable first.

This is one example but not the only one.

Enselic:
This will mean that e.g. rustup-toolchain can depend on `public-api` that is older than what we have in our workspace, but that's fine. It only depends on `public-api` so that we can make sure to not break its public API. It doesn't need the latest version for that. That is the same situation we would have it it lived in a separate git repo, which is kind of what we are simulating.

Emilgardis:
I see, the problem then is dev dependencies. Not "real" dependencies.

Wouldny it be better to keep it as one workspace but have the tests not create circular behaviour by using tests in another crate/package

Enselic:
I experimented a bit in that direction before this rewrite, and it didn't seem possible to get a nice solution with both a crates.io version and a local version of the same crate in the same workspace.

Do you have any particular concerns with this new setup? Hopefully I can address or mitigate all of your concerns.

Emilgardis:
my concern is that there's two workspaces in one git tree, and the reason for it is unnecessary. 

the test in [`rustup-toolchain`](https://github.com/cargo-public-api/cargo-public-api/blob/ceee4d03309ea8479fbe630fc73485c98abbc04b/rustup-toolchain/tests/rustup-toolchain-lib-tests.rs) is redundant when considering the rest of the workspace. The same goes for the tests in `rustdoc-json` that use `public-api`. The tests should in my opinion test the surface of the crate itself, not how other crates work (and this becomes especially true when it creates circular dependencies).

I propose we remove the dev-dependencies that causes this and instead, if there's a test missing, introduce them in either `public-api` or create a new workspace member that is only used for testing.

Enselic:
Hmm interesting idea, I'll look into it.

> The tests should in my opinion test the surface of the crate itself

Just to make sure we don't talk past each other: the only purpose of those tests _is_ to test the API surface of those crates. But I agree that in our setup it can make sense to do that in another package just to avoid circular dependencies.

Emilgardis:
> Just to make sure we don't talk past each other: the only purpose of those tests _is_ to test the API surface of those crates.

Yes of course! worded it badly. The tests use a dependency they don't strictly need to do the test of the API surface - was what I tried (and failed) to convey.

Enselic:
I'll close this for now because I don't see why we need it open any longer. Thanks for review!

PR #706, opened 2025-02-10
Avoid circular dependencies in workspace (move public-api tests)

Enselic:
Right now there are circular dependencies in the workspace, which makes making releases feel like a minefield. Making releases must be effortless and fun, so let's stop having circular dependencies.

Break the circular dependency by moving the public API surface tests upstream. Idea from Emil.

The public API seems stable enough to stop worrying about keeping docs in sync with code.

Alternative to https://github.com/cargo-public-api/cargo-public-api/pull/702


Enselic:
I'll merge this and try it out with the upcoming release

PR #710, opened 2025-02-19
workflows: Stop using Swatinem/rust-cache

Enselic:
To reduce risk of cache poisoning pwn requests.
See e.g.
https://securitylab.github.com/resources/github-actions-new-patterns-and-mitigations/

This will lead to slower CI but I prefer that over being paranoid about potential cache poisoning and other exploits.

If we end up hating the slower CI we can look into alternatives, but let's at least try.

### DIFF
--- a/.github/workflows/Nightly.yml
+++ b/.github/workflows/Nightly.yml
@@ -46,7 +46,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: rustup install nightly --profile minimal
-      - uses: Swatinem/rust-cache@v2
       - run: sudo apt-get install -y zsh && zsh --version
       - run: ./scripts/cargo-test.sh --bless
       - id: latest-nightly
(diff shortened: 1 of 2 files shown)


.github/workflows/Nightly.yml
```diff
@@ -46,7 +46,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: rustup install nightly --profile minimal
-      - uses: Swatinem/rust-cache@v2
```
Emilgardis, line 49: how is the linked article related to this workflow? this runs on branches on this repo only

Enselic:
Looks like the only slow job is the Ubuntu job that uses the legacy git crates.io protocol. The sparse crates.io protocol has been the default for 1,5 years IIRC, so I think we can simply stop testing the git protocol, but keep the code. Once we get  breakage we remove the code as well and stop supporting it. Most probably no one is using that any longer.

Emilgardis (commented):
the article mentions that cache posioning happens in `pull_request_target`, not `pull_request`. Not sure I see how this applies

.github/workflows/Nightly.yml
Enselic: right, this is more of a move in the general direction of "stop running third party code" and "reduce exposure to risks"

For example, we don't have to worry about rust-cache itself becoming compromised. It does run in a context with one of our secrets after all.

PR #712, opened 2025-02-20
Auto-merge-dependabot-PRs.yml: Do not evaulate context expressions in shell context

Enselic:
Auto-merge-dependabot-PRs.yml is our only `pull_request_target` triggered workflow and that type is extra risky.

From
https://securitylab.github.com/resources/github-actions-untrusted-input/ we can read:

> The best practice to avoid code and command
> injection vulnerabilities in GitHub workflows is
> to set the untrusted input value of the
> expression to an intermediate environment
> variable

And even though `${{ github.event.pull_request.number }}` is not free-text user controlled, there might be a bug in GitHub itself that allows free text in that field today or in the future. So protect us from pwn requests even in that unlikely scenario by putting the PR number in an env var.

### DIFF
--- a/.github/workflows/Auto-merge-dependabot-PRs.yml
+++ b/.github/workflows/Auto-merge-dependabot-PRs.yml
@@ -15,16 +15,19 @@ jobs:
       url: https://github.com/cargo-public-api/cargo-public-api/blob/main/.github/workflows/Auto-merge-dependabot-PRs.yml
     env:
       GITHUB_TOKEN: ${{ secrets.AUTO_MERGE_GITHUB_TOKEN }}
+      PR: ${{ github.event.pull_request.number }}
     steps:
       - run: |
-          gh pr review ${{ github.event.pull_request.number }} \
+          gh pr review \
             --repo cargo-public-api/cargo-public-api \
             --comment \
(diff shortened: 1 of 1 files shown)


.github/workflows/Auto-merge-dependabot-PRs.yml
```diff
@@ -15,16 +15,19 @@ jobs:
...
+      PR: ${{ github.event.pull_request.number }}
     steps:
       - run: |
-          gh pr review ${{ github.event.pull_request.number }} \
+          gh pr review \
             --repo cargo-public-api/cargo-public-api \
             --comment \
-            --body "If CI passes, this dependabot PR will be [auto-merged](https://github.com/cargo-public-api/cargo-public-api/blob/main/.github/workflows/Auto-merge-dependabot-PRs.yml) 🚀"
+            --body "If CI passes, this dependabot PR will be [auto-merged](https://github.com/cargo-public-api/cargo-public-api/blob/main/.github/workflows/Auto-merge-dependabot-PRs.yml) 🚀" \
+            '$PR'
```
smoelius: Should this and line 31 be `"$PR"`? (Sorry to intrude, BTW.)

Enselic: Oops, good catch, thanks!

I feel nothing but gratitude whenever someone comes along and helps with code review, so feel free to continue doing that  : )

PR #715, opened 2025-02-21
Support `use<..>` bound

### LINKED ISSUE #714: Support `use<..>` generic bound
I encountered the following panic when running `cargo public-api` v0.44.0 on [`clippy_utils`](https://github.com/rust-lang/rust-clippy/tree/238edf273d195c8e472851ebd60571f77f978ac8/clippy_utils) with the `238edf273d195c8e472851ebd60571f77f978ac8` commit. Note that `clippy_utils` is nightly-only, but I believe `use<..>` has been stabilized:

```bash
$ RUST_BACKTRACE=1 cargo +nightly-2025-02-20 public-api
thread 'main' panicked at /Users/USER/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/public-api-0.44.0/src/render.rs:953:37:
not yet implemented: Can this code path be triggered with just stable Rust? If you hit this todo with just stable Rust, please file an issue with a minimal reproducer. Thanks!
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: public_api::render::RenderingContext::render_sequence_impl
   3: public_api::render::RenderingContext::render_option_type
   4: public_api::render::RenderingContext::render_fn_decl
   5: public_api::render::RenderingContext::token_stream
   6: public_api::public_item::PublicItem::from_intermediate_public_item
   7: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   8: public_api::item_processor::public_api_in_crate
   9: public_api::Builder::build
  10: cargo_public_api::api_source::public_api_from_rustdoc_json
  11: <cargo_public_api::api_source::CurrentDir as cargo_public_api::api_source::ApiSource>::obtain_api
  12: cargo_public_api::main_
  13: cargo_public_api::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I'm running MacOS on an M1 chip, but that shouldn't affect the error. The panicking line of code is here:

https://github.com/cargo-public-api/cargo-public-api/blob/124bdfd85e46a68088d6b0009682c101534f3664/public-api/src/render.rs#L953

I'm still working on making a smaller reproducible example, I'll update this issue when I have one!

BD103:
Fixes #714, please see it for more information!

With this PR, the following code is now processed without crashing:

```rust
pub fn capture<'a, 'b, T>(x: &'a (), y: T) -> impl Sized + use<'a, T> {
    //                                        ~~~~~~~~~~~~~~~~~~~~~~~
    //                                       Captures `'a` and `T` only.
    (x, y)
}
```

This is my first time contributing! Where can I add tests for my changes? Is there anything else I need to do?

Thanks! :)

### DIFF
--- a/cargo-public-api/tests/snapshots/cargo_public_api_bin_tests__comprehensive_api.snap
+++ b/cargo-public-api/tests/snapshots/cargo_public_api_bin_tests__comprehensive_api.snap
@@ -117,6 +117,8 @@ pub fn comprehensive_api::functions::return_tuple() -> (bool, comprehensive_api:
 pub fn comprehensive_api::functions::somewhere<T, U>(t: T, u: U) where T: core::fmt::Display, U: core::fmt::Debug
 pub fn comprehensive_api::functions::struct_arg(s: comprehensive_api::structs::PrivateField)
 pub fn comprehensive_api::functions::synthetic_arg(t: impl comprehensive_api::traits::Simple) -> impl comprehensive_api::traits::Simple
+pub fn comprehensive_api::functions::synthetic_use_capture<'a, 'b, T>(x: &'a (), y: T) -> impl core::marker::Sized + use<'a, T>
+pub fn comprehensive_api::functions::synthetic_use_no_capture<'a>(x: &'a usize) -> impl core::marker::Sized + use<>
 pub unsafe fn comprehensive_api::functions::unsafe_fn()
 pub fn comprehensive_api::functions::unused_argument(_: u32)
 pub mod comprehensive_api::higher_ranked_trait_bounds
--- a/public-api/src/render.rs
+++ b/public-api/src/render.rs
@@ -941,7 +941,7 @@ impl<'c> RenderingContext<'c> {
                 let mut output = vec![];
                 output.extend(self.render_higher_rank_trait_bounds(generic_params));
                 match modifier {
-                    TraitBoundModifier::None | TraitBoundModifier::MaybeConst=> {}
+                    TraitBoundModifier::None | TraitBoundModifier::MaybeConst => {}
                     TraitBoundModifier::Maybe => {
                         output.push(Token::symbol("?"));
                     }
--- a/public-api/src/render.rs
+++ b/public-api/src/render.rs
@@ -950,7 +950,30 @@ impl<'c> RenderingContext<'c> {
                 output
             }
             GenericBound::Outlives(id) => vec![Token::lifetime(id)],
-            GenericBound::Use(_) => todo!("Can this code path be triggered with just stable Rust? If you hit this todo with just stable Rust, please file an issue with a minimal reproducer. Thanks!"),
+            GenericBound::Use(params) => {
+                let mut output = vec![Token::keyword("use"), Token::symbol("<")];
+
+                for i in 0..params.len() {
+                    let param = &params[i];
+
+                    // Rustdoc doesn't tell us if the parameter is a lifetime or a generic, so we
+                    // check if the string starts with `'` to see if it is a lifetime.
+                    if param.starts_with('\'') {
+                        output.push(Token::lifetime(param));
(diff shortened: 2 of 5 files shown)


Enselic:
Thanks! As a first step, please add a reproducer to https://github.com/cargo-public-api/cargo-public-api/blob/main/test-apis/comprehensive_api/src/functions.rs, run ./scripts/cargo-test.sh --bless, and push the resulting commit here so we can see how the rendering looks, and also prevent regressions.

public-api/src/render.rs
```diff
@@ -941,7 +941,7 @@ impl<'c> RenderingContext<'c> {
                 let mut output = vec![];
                 output.extend(self.render_higher_rank_trait_bounds(generic_params));
                 match modifier {
-                    TraitBoundModifier::None | TraitBoundModifier::MaybeConst=> {}
+                    TraitBoundModifier::None | TraitBoundModifier::MaybeConst => {}
```
Enselic, line 944: (We run rustfmt in  CI so it is strange that you got formatting problems)

BD103: Yeah, I'm surprised this wasn't caught earlier! I did run `cargo fmt --all` instead of the normal `cargo fmt`, so that may have caused it.

BD103:
Looking at `lint.sh`, I don't think it runs `rustfmt` on crates outside of the workspace:

https://github.com/cargo-public-api/cargo-public-api/blob/124bdfd85e46a68088d6b0009682c101534f3664/scripts/lint.sh#L4

You could switch it to `cargo fmt --check --all`, which I think would help format examples and tests, but it still wouldn't format these crates:

https://github.com/cargo-public-api/cargo-public-api/blob/124bdfd85e46a68088d6b0009682c101534f3664/Cargo.toml#L32-L71

cargo-public-api/tests/snapshots/cargo_public_api_bin_tests__comprehensive_api.snap
```diff
@@ -117,6 +117,8 @@ pub fn comprehensive_api::functions::return_tuple() -> (bool, comprehensive_api:
 pub fn comprehensive_api::functions::somewhere<T, U>(t: T, u: U) where T: core::fmt::Display, U: core::fmt::Debug
 pub fn comprehensive_api::functions::struct_arg(s: comprehensive_api::structs::PrivateField)
 pub fn comprehensive_api::functions::synthetic_arg(t: impl comprehensive_api::traits::Simple) -> impl comprehensive_api::traits::Simple
+pub fn comprehensive_api::functions::synthetic_use_capture<'a, 'b, T>(x: &'a (), y: T) -> impl core::marker::Sized + use<'a, T>
```
Enselic, line 120: This is rendered in the same way cargo doc renders it :+1: 

image

cargo-public-api/tests/snapshots/cargo_public_api_bin_tests__comprehensive_api.snap
```diff
@@ -117,6 +117,8 @@ pub fn comprehensive_api::functions::return_tuple() -> (bool, comprehensive_api:
...
+pub fn comprehensive_api::functions::synthetic_use_no_capture<'a>(x: &'a usize) -> impl core::marker::Sized + use<>
```
Enselic, line 121: and to my surprise, this too, despite the strange looking `use<>`:

image

public-api/src/render.rs
Enselic: playing around a bit with it it seems like a cargo fmt bug for this particular closure code :shrug:

Enselic (approved):
Looks good, thanks!

I'll merge this now and will likely make a release later today.

BD103:
Thank you! :)

PR #722, opened 2025-02-25
Stop depending on cargo-insta binary, use INSTA_UPDATE instead

Enselic:
Because that is simpler.


Enselic:
Merging to avoid conflict with upcoming PR

PR #723, opened 2025-02-25
Replace peter-evans/create-pull-request with custom script

Enselic:
To reduce security risk since the job in question
has access to the token of the `EnselicCICD` user.

~~Todo~~ Done:
- [x] Verify, which will require me to spam PRs for a while, sorry about that.

### DIFF
--- a/scripts/create-auto-bless-pr.sh
+++ b/scripts/create-auto-bless-pr.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+set -o nounset -o errexit -o pipefail
+
+# When this script is run there are already local git repo changes with new
+# blessed output using latest nightly toolchain. Create a branch to work with.
+nightly_version="nightly-$(date +'%Y-%m-%d')"
+branch_name="auto-bless/$nightly_version"
+git checkout \
+    -b "$branch_name"
+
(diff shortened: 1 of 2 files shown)


scripts/create-auto-bless-pr.sh
```diff
@@ -0,0 +1,28 @@
...
+nightly_version="nightly-$(date +'%Y-%m-%d')"
+branch_name="auto-bless/$nightly_version"
+git checkout \
+    -b "$branch_name"
+
+# Update MINIMUM_NIGHTLY_RUST_VERSION_FOR_TESTS and then commit all changes.
+echo $nightly_version >cargo-public-api/MINIMUM_NIGHTLY_RUST_VERSION_FOR_TESTS
+git add .
+git commit \
+    --author "EnselicCICD <junta-pixlar0l@icloud.com>" \
```
Emilgardis: could use `104096785+EnselicCICD@users.noreply.github.com` here to not expose this email in the repo ;)

Enselic: thanks, good catch : )

(old commits seems to use the real email sometimes  but I agree it is unnecessary to expose it in the versioned files themselves)

PR #735, opened 2025-03-04
Debug cargo missing in CI

Enselic:
Debugs https://github.com/cargo-public-api/cargo-public-api/issues/734


Enselic:
See https://github.com/cargo-public-api/cargo-public-api/issues/734#issuecomment-2696281497 for conclusion

PR #757, opened 2025-04-18
Revert "Bring back dependabot auto-merge workflow but don't use `pull_request_target`"

Enselic:
That didn't work. I have created a PR to improve GithUb docs: https://github.com/github/docs/pull/37635


Emilgardis:
Not sure I understand the reasoning here, this works just fine, looks to me like the merge queue is just misconfigured. ~~There's no CI for the branches the merge queue creates~~ (this is wrong, missed that we use `merge_group`)

Enselic:
(I'll reply in detail when I got time, currently busy)

Maybe your right, I didn't look super close

Enselic:
Ok I see you edited

I think the problem is that github token don't trigger workflows 

Link in pr I linked to

PR #776, opened 2025-05-12
Stop using EnselicCICD crates.io token for publish

Enselic:
The reason we use EnselicCICD in a GitHub Org Team to make releases is that before https://github.com/rust-lang/crates.io/issues/5443 was implemented, that was the only way to have a crates.io token that did not allow changing owners, which is needed to avoid hostile takeovers in case of token leaks.

Now that https://github.com/rust-lang/crates.io/issues/5443 has been implemented for quite a long time, we can switch over to using `publish-update`-scoped tokens (i.e. tokens that do not have the `change-owners` scope).

For now I just added my token, but other maintainers are free to add their tokens for when they make releases. We'd need to figure out exactly how to safetly make a toggle, however.

After this transition is complete, I think we can remove EnselicCICD from our org, which would strengthen our security. Not that that account has low security, but it's just an unnecessary attack surface.

### DIFF
--- a/.github/workflows/Release-all.yml
+++ b/.github/workflows/Release-all.yml
@@ -27,7 +27,7 @@ jobs:
   release-rustup-toolchain:
     if: ${{ !cancelled() && inputs.release_rustup_toolchain }}
     permissions:
-      contents: write # git push + access to secrets.CARGO_REGISTRY_TOKEN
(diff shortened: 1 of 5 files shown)


Enselic:
For the record, we still will use EnselicCICD for auto-bless, but that should not require membership in our org.

PR #785, opened 2025-05-30
Fix: make PR #783 compile (type mismatch in library_name)

Enselic:
This PR fixes a type mismatch in rustdoc-json/src/builder.rs that prevented PR #783 from compiling. The comparison and return of package names now use String as expected.


Enselic:
Sigh, so many new dependencies. I'll pass this one.

PR #804, opened 2025-07-13
Prepare for trusted publishing to crates.io

Enselic:
See https://crates.io/docs/trusted-publishing.

That way we don't need to store a long-lived token in our repository settings. The only secret someone needs to store the JWT private key of GitHub. I think it is safe to assume that their security engineers are going to do a better job at keeping secrets than us.

TODO:
- [x] Wait for a need to make a release
- [x] Configure Trusted Publishing on crates.io
- [x] Verify this PR by making the release(s) through Trusted Publishing
- [ ] Remove the token from our repo secrets once we have confirmed this works

### DIFF
--- a/.github/workflows/Release-all.yml
+++ b/.github/workflows/Release-all.yml
@@ -27,30 +27,30 @@ jobs:
   release-rustup-toolchain:
     if: ${{ !cancelled() && inputs.release_rustup_toolchain }}
     permissions:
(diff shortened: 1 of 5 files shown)


Enselic:
Everything seems to work now. The publisher is "Via GitHub":

image

I had to add separate configurations for Release-all.yml and the individual release scripts, but that's fine.

PR #809, opened 2025-07-17
CI: Windows does not run any tests, so remove

Enselic:
Windows does not actually run any tests so CI is always green.

Let's remove the job to not make it appear like we test for Windows, until someone has interest and time to fix the Windows CI.

cc @Emilgardis who maybe want to take a look? IIRC you added the Windows job

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -59,7 +59,6 @@ jobs:
         variant:
           - { name: Ubuntu,  os: ubuntu-latest  }
           - { name: macOS,   os: macos-latest   }
-          - { name: Windows, os: windows-latest }
     name: cargo test (${{ matrix.variant.name }})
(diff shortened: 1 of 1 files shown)


Enselic:
example of where Windows should have failed but didn't (since no tests ran): https://github.com/cargo-public-api/cargo-public-api/actions/runs/16335996724

Enselic:
Oops, didn't mean to add Copilot as reviewer...

Emilgardis:
why does it not run tests?

Emilgardis:
ah I see, we just need to specify shell: bash for the run command and it should be fixed

Enselic:
If you know what to do maybe you could try, either in a new PR or this one? I suspect some tests will fail.

Enselic:
obsoleted by https://github.com/cargo-public-api/cargo-public-api/pull/811

PR #830, opened 2025-09-15
Add support for passing environment variables to the child process

vvvar:
Hi! 👋

First of all, thank you for this project!

I ran into a use case where I need `rustdoc-json` as a library in multi-threaded code.
I need to pass an environment variables to the `rustc` that is running in the child process spawned by `rustdoc-json`.
The problem is that using `std::env::set_var` is not an option for me because it is unsafe in a multi-threaded code.
Setting environment variables in the `shell` is not feasible for me as well.

This PR is a proposal to add support for passing environment variables to the spawned child process.
I hope this could be useful to others as well.

### What?

- Add Builder::env(key, val) (mirroring std::process::Command::env).
- Update snapshot.
- Add a test.

Please let me know are you interested in this contribution.
If you’re open to it, I’ll be happy to make any adjustments needed to get this merged.

### DIFF
--- a/rustdoc-json/src/builder.rs
+++ b/rustdoc-json/src/builder.rs
@@ -2,6 +2,8 @@ use super::BuildError;
 use cargo_metadata::TargetKind;
 use tracing::*;
 
+use std::collections::HashMap;
+use std::ffi::{OsStr, OsString};
 use std::io::Write;
 use std::{
     path::{Path, PathBuf},
@@ -89,6 +91,7 @@ fn cargo_rustdoc_command(options: &Builder) -> Result<Command, BuildError> {
         package_target,
         document_private_items,
         cap_lints,
+        envs,
     } = options;
 
     let mut command = match OVERRIDDEN_TOOLCHAIN.or(requested_toolchain.as_deref()) {
@@ -156,6 +159,7 @@ fn cargo_rustdoc_command(options: &Builder) -> Result<Command, BuildError> {
     if let Some(cap_lints) = cap_lints {
         command.args(["--cap-lints", cap_lints]);
     }
+    command.envs(envs);
     Ok(command)
 }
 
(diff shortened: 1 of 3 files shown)


Enselic:
Hi! I took a quick look and this looks very reasonable, thanks! I'll give some time for other maintainers to comment if they want to, but I currently see no blockers for merging this.

rustdoc-json/src/builder.rs
```diff
@@ -431,6 +437,24 @@ impl Builder {
...
 
+    /// Environment variable mapping to pass to the spawned `cargo rustdoc` process.
+    ///
+    /// # Notes
+    ///
+    /// - Environment variable names are case-insensitive (but case-preserving) on Windows and case-sensitive on all other platforms.
+    /// - Spawned `cargo rustdoc` processes will inherit environment variables from their parent process by default.
+    ///   Environment variables explicitly set using this take precedence over inherited variables.
+    #[must_use]
+    pub fn env<K, V>(mut self, key: K, val: V) -> Self
```
Enselic: Can you change to `key: impl AsRef<OsStr>, val: impl AsRef<OsStr>` please? To follow the style in the rest of the file (it does not affect the public API, it's just a cleaner syntax).

Enselic (changes requested):
Looks good! Just had a nit. After that is fixed I'm happy to merge.

rustdoc-json/src/builder.rs
Enselic: @vvvar Is this PR still needed?

vvvar: Sure. Updated. Sorry for the long response.

Enselic (approved):
thanks

Enselic:
This will be merged soon.

I guess there is no rush to get a release out, so I'll see when I get to it.

If you want to help it get done faster you could look at https://github.com/cargo-public-api/cargo-public-api/blob/main/docs/RELEASE.md and prepare a PR for it. Thanks in advance if you're interested!

It's probably enough just to do the steps in https://github.com/cargo-public-api/cargo-public-api/blob/main/docs/RELEASE.md#rustdoc-json.

PR #831, opened 2025-09-16
`Box<dyn ...>` with two or more traits

smoelius:
This is more of an issue than a PR.

If you have a `Box<dyn ...>` type with two or more traits, `public-api` wants to wrap the `dyn ...` in parentheses.

However, nightly Rust then complains about the parentheses ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=85c40071d77d08c5c104fec63f065562)).

The problem appears to be here: https://github.com/cargo-public-api/cargo-public-api/blob/0d022e13912af7036340729b2df254c5ef467ad6/public-api/src/render.rs#L423-L426

I don't know how easy this would be to fix.

### DIFF
--- a/cargo-public-api/tests/snapshots/comprehensive_api.txt
+++ b/cargo-public-api/tests/snapshots/comprehensive_api.txt
@@ -93,6 +93,7 @@ pub fn comprehensive_api::exports::v1::foo2()
 pub mod comprehensive_api::functions
 pub async fn comprehensive_api::functions::async_fn()
 pub async fn comprehensive_api::functions::async_fn_ret_bool() -> bool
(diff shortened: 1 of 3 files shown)


Enselic (approved):
Thanks for the bug report. The clarity of doing it by adding a test case with a PR can't be topped 👍 

By my estimation this is a quite small problem though? So I suspect no one will be bothered enough by it to fix it?

So I suggest we simply merge this PR so that we have a test in place if someone comes around to fix it.

We could report a regular issue for visibility, but by my estimation that would just be administrative overhead. Most likely no one will fix the issue and I will end up closing as "won't fix" to keep the issue inbox clean.

Let me know what you think!

smoelius:
Your plan sounds fine to me. Thanks a lot for looking at this.

Enselic (approved):
👍merging

PR #869, opened 2026-02-23
Add support for including trait default functions

gatesn:
Currently if an implementation overrides a defaulted trait function the public API appears to have changed. In practice, the user facing API has always had this function on it.

### DIFF
--- a/public-api/src/item_processor.rs
+++ b/public-api/src/item_processor.rs
@@ -240,10 +240,10 @@ impl<'c> ItemProcessor<'c> {
     ) {
         let finished_item = unprocessed_item.finish(item, overridden_name, type_);
 
-        let children = children_for_item(item).into_iter().flatten();
+        let children = self.children_for_item(item);
         let impls = impls_for_item(item).into_iter().flatten();
 
-        for &id in children {
+        for &id in &children {
             self.add_to_work_queue(finished_item.path().into(), Some(item.id), id);
         }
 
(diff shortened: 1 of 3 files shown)


Enselic (changes requested):
Hi, let's split this up to make review easier. First, please create a separate PR that adds traits with and without (if applicable) default impls to `test-apis/comprehensive_api/src/traits.rs`. And make CI pass. Without the fix.

Then rebase this PR on top. That makes it easy to see the impact the fix has on how the public API is presented.

I also don't think this needs to be opt-in, i.e. I don't think we need a new --include? Why would someone not want your fix?

Also please follow the steps at `public-api/CHANGELOG.md`. We will bump to 0.51.0.

gatesn:
Sounds like a plan!

I've opened the first PR here: https://github.com/cargo-public-api/cargo-public-api/pull/871

Once merged I'll follow up with a PR to alter the behavior. In terms of the release, do you want that as part of the second PR also? Or you prefer a third PR to bump version numbers + changelog (I'm guessing per https://github.com/cargo-public-api/cargo-public-api/blob/main/docs/RELEASE.md ?)

Enselic:
Yeah let's do a third PR after this PR to prepare for releasing according to RELEASE.md (I mis-linked it, so I'm glad you found it), thanks.

public-api/tests/public-api-lib-tests.rs
```diff
@@ -239,6 +239,54 @@ fn auto_traits() {
     );
 }
 
+#[test]
```
Enselic: this tests the same thing we already test with comprehensive_api I think? I don't think the comprehensive_api test can pass while this test fails (and catches regressions that comprehensive_api didn't)? We can probably remove this test?

public-api/src/item_processor.rs
```diff
@@ -191,6 +191,40 @@ impl<'c> ItemProcessor<'c> {
...
 
         self.process_item_for_type(unprocessed_item, item, None, Some(&impl_.for_));
+
+        // For trait impls, include default trait items that were not
+        // overridden. This ensures that overriding a default method does
+        // not cause a spurious diff in the public API output.
+        if let Some(trait_path) = &impl_.trait_ {
+            // Get the path from the impl item we just pushed to output.
+            let impl_path: Vec<PathComponent<'c>> =
+                self.output.last().expect("just pushed").path().to_vec();
```
Enselic: `output.last()` seems quite hacky.

this logic can probably be greatly simplified if you do it in `fn children_for_item()` instead? Or maybe some other place. I didn't look very closely. At least try to find the proper place "in the middle of the action" rather than retroactively "after everything already happened".

Enselic (changes requested):
Thank you for the update. Here is the first round of comments.

Also, can you follow https://github.com/cargo-public-api/cargo-public-api/blob/028fb951e8b242b0dcfa105effe5c1c1f7925d3b/docs/CONTRIBUTING.md#creating-git-commits please? That is a new section not even merged yet (see https://github.com/cargo-public-api/cargo-public-api/pull/873).  But I'd like to avoid 10 commits. This PR should be one commit, to play well with git tooling such as `git bisect` etc etc.

public-api/src/item_processor.rs
gatesn: Yeah agreed. I took another stab, but it means returning an owned Vec from children_for_item, instead of a ref. Not sure if you have benchmarks or are concerned about this performance wise at all.

Enselic: I benchmarked your code and real-life performance only seems to have regressed with ~1%. I think we can live with that, even if the code looks optimizable. We can always optimize later if we want. The most important thing is to not have bugs.

Here are the benchmark numbers. I used our very ad-hoc benchmark rig found [here](https://github.com/cargo-public-api/cargo-public-api-benchmarks/blob/10f89b263cdbcd7db8c69a3f03f7a9cb76ba563d/benches/json_processing.rs#L21-L42).

Without your fix:
```
$ cargo bench --  aws_sdk_ec2
process aws_sdk_ec2 JSON
                        time:   [4.1068 s 4.1257 s 4.1440 s]
```

with your fix:

```
$ cargo bench --  aws_sdk_ec2
process aws_sdk_ec2 JSON
                        time:   [4.1385 s 4.1744 s 4.2103 s]
                        change: [+0.2462% +1.1808% +2.1949%] (p = 0.03 < 0.05)
```

Enselic (approved):
Thanks, let's merge this.

For the next PR, you only need to prepare `public-api` and `cargo-public-api` crates for a release.  We will do a 0.x.0 bump to 0.51.0.

(I will take care of the `rustup-toolchain` and `rustdoc-json` crates.)

Here is a proposal for a changelog entry:
* Also include default trait methods in `impl`s, even if not overridden, just like in rustdoc HTML.

PR #871, opened 2026-02-24
Add traits with default methods to comprehensive_api test suite

gatesn:
Add TraitWithDefaultMethod with required and default methods, plus three structs (UsesAllDefaults, OverridesSomeDefaults, OverridesAllDefaults) that exercise different override patterns.

The API lock file shows only the overridden methods are included, not the defaulted ones. A follow up PR will fix this to include the defaulted functions.

### DIFF
--- a/test-apis/comprehensive_api/src/impls.rs
+++ b/test-apis/comprehensive_api/src/impls.rs
@@ -1,7 +1,7 @@
 use crate::{
     structs::{Plain, Unit, WithLifetimeAndGenericParam},
     traits::{
-        GenericAssociatedTypes, Simple, TraitReferencingOwnAssociatedType, TraitWithDefaultMethod,
+        GenericAssociatedTypes, Simple, TraitReferencingOwnAssociatedType, TraitWithOneRequiredAndTwoDefaultMethods,
         TraitWithGenerics,
     },
 };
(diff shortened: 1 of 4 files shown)


test-apis/comprehensive_api/src/traits.rs
```diff
@@ -57,3 +57,15 @@ pub trait GenericAssociatedTypes {
     type SimpleBound: Simple;
     type WithLifetime<'a>;
 }
+
+pub trait TraitWithDefaultMethod {
```
Enselic: let's have a name that is as clear as possible, so one can see what it does without looking at its code
```suggestion
pub trait TraitWithOneRequiredAndTwoDefaultMethods {
```

test-apis/comprehensive_api/src/impls.rs
```diff
@@ -117,6 +118,45 @@ impl GenericAssociatedTypes for Unit {
...
+        false
+    }
+
+    fn defaulted(&self) -> u32 {
+        99
+    }
+}
+
+/// Overrides all methods, no defaults used.
+pub struct OverridesAllDefaults;
```
Enselic: How about renaming to

```suggestion
pub struct OverridesRequiredAndBothDefaults;
```

? Since it overrides all methods, not only the default ones.

test-apis/comprehensive_api/src/impls.rs
```diff
@@ -117,6 +118,45 @@ impl GenericAssociatedTypes for Unit {
...
+pub struct UsesAllDefaults;
+
+impl TraitWithDefaultMethod for UsesAllDefaults {
+    fn required(&self) -> bool {
+        true
+    }
+}
+
+/// Overrides `required` and `defaulted`, but leaves `another_default`.
+pub struct OverridesSomeDefaults;
```
Enselic:
```suggestion
pub struct OverridesRequiredAndOneOfTwoDefaults;
```

test-apis/comprehensive_api/src/impls.rs
```diff
@@ -117,6 +118,45 @@ impl GenericAssociatedTypes for Unit {
     type WithLifetime<'a> = GatTestStruct1<'a, bool>;
 }
 
+/// Only overrides `required`, leaving both defaults.
+pub struct UsesAllDefaults;
```
Enselic:
```suggestion
pub struct OverridesOnlyRequired;
```

Enselic (changes requested):
Thanks, looks good. I would just like the names to be even clearer.

For the record, rustdoc HTML also renders default methods even if not overridden, so according to [our own rules](https://github.com/cargo-public-api/cargo-public-api/blob/main/docs/CONTRIBUTING.md#expected-output) we are clearly buggy in this regard:

Image

Enselic (approved):
thanks

PR #874, opened 2026-02-26
Bump cargo-public-api and public-api to 0.51.0 for release

gatesn:
I didn't know if you wanted me to bump nightly rust version also. Seems like the only release that hasn't given the formatting in the README.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # `cargo-public-api` changelog
 
+## v0.51.0
+* Also include default trait methods in impls, even if not overridden, just like in rustdoc HTML.
+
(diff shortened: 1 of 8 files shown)


Enselic (approved):
Everything looks correct, thanks.

You were right to not bump nightly version, since the rustdoc JSON format has not changed. But since we change how items are rendered, we can't just do a 0.0.x bump since that would break peoples CI tests. That's why we need to bump 0.x.0.

I'll merge this and a couple of other PRs and then trigger the release jobs.

gatesn:
Great! Thank you for shepherding this over the line. It's been a great project for us 🙏

PR #886, opened 2026-04-30
feat: optionally omit parameter names flag

joseph-isaacs:
I have just added the option to remove parameter names from the publicly generated API files, I was wondering if this would be something you want to upstream.

## Motivation

Our usage of this is to store the public api for all crates in our project and we found that having the names of public method cases extra noise. Renaming or not using a variable doesn't change the method signature.

## Usage

```rust
let builder = MAKE_BUILDER
  .omit_param_names(true)
```

relevant issue: https://github.com/cargo-public-api/cargo-public-api/issues/766

### DIFF
--- a/public-api/src/lib.rs
+++ b/public-api/src/lib.rs
@@ -73,6 +73,7 @@ struct BuilderOptions {
     omit_blanket_impls: bool,
     omit_auto_trait_impls: bool,
     omit_auto_derived_impls: bool,
+    omit_param_names: bool,
 }
 
 /// Builds [`PublicApi`]s. See the [top level][`crate`] module docs for example
@@ -94,6 +95,7 @@ impl Builder {
             omit_blanket_impls: false,
             omit_auto_trait_impls: false,
             omit_auto_derived_impls: false,
+            omit_param_names: false,
         };
         Self {
             rustdoc_json: path.into(),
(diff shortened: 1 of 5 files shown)


public-api/src/lib.rs
```diff
@@ -167,6 +169,23 @@ impl Builder {
         self
     }
 
+    /// If `false` (default), function parameter names are contained in the output.
```
Enselic, line 172:
```suggestion
    /// If `false` (default), function parameter names are included in the output.
```

I don't speak English natively, but doesn't "included" instead of "contained" sound better?

public-api/src/lib.rs
```diff
@@ -167,6 +169,23 @@ impl Builder {
...
+    /// pub fn public_api::Builder::omit_param_names(self, omit_param_names: bool) -> Self
+    /// ```
+    ///
+    /// If `true`, function parameter names are omitted from the output.
+    ///
+    /// ```
+    //  pub fn public_api::Builder::omit_param_names(self, bool) -> Self
+    //  ```
+    #[must_use]
+    pub fn omit_param_names(mut self, omit_param_names: bool) -> Self {
```
Enselic, line 184: I personally prefer clarity over terseness in public APIs. What do you think about calling this `omit_fn_parameter_names()`?

Enselic (changes requested):
Unless other maintainers raise objections, I'm fine with this addition.

Enselic:
After thinking some more about this, I wonder if we maybe should hide parameter names by default, and not even have an opt in. If users hate it, we can always add an opt-in in a 0.0.x release.

Enselic:
See https://github.com/cargo-public-api/cargo-public-api/pull/888

Enselic:
(See https://github.com/cargo-public-api/cargo-public-api/pull/888)

PR #895, opened 2026-05-27
Re-export `rustdoc-types`

### LINKED ISSUE #794: Consider reexporting rustdoc-types dependency
Hello,

This crate is very useful, especially the formatting and sorting of items. So thanks!

I was wondering if you would consider re-exporting the rustdoc-types version (ie `pub use rustdoc_types;`), so it is easier for downstream users to take [`PublicItem::id()`](https://docs.rs/public-api/latest/public_api/struct.PublicItem.html#method.id) and look it up in a `rustdoc_types::Crate` without crate version mismatches.

Thanks!

smoelius:
Re-export `rustdoc-types` under the feature `experimental-feature-that-can-be-removed-in-a-patch-release_re-export-rustdoc-types`.

Resolves #794.

Nits are welcome.

### DIFF
--- a/public-api/Cargo.toml
+++ b/public-api/Cargo.toml
@@ -13,6 +13,8 @@ repository = "https://github.com/cargo-public-api/cargo-public-api/tree/main/pub
 [features]
 default = ["snapshot-testing"]
 snapshot-testing = ["dep:snapshot-testing"]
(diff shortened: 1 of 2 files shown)


Enselic (approved):
LGTM

To get a release out faster, consider doing the steps at https://github.com/cargo-public-api/cargo-public-api/blob/main/docs/RELEASE.md (you only need to follow the steps for the public-api crate)

PR #896, opened 2026-05-29
Bump version

smoelius:
Hopefully, I did this right (ref: https://github.com/cargo-public-api/cargo-public-api/pull/895#pullrequestreview-4386485322). I assume a patch release is okay(?).

Aside, in the following link, I think `lib` should be `src`: https://github.com/cargo-public-api/cargo-public-api/blob/cafae1356b2fccb7e9c2250fced0aa71d4281f99/docs/RELEASE.md?plain=1#L67

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -903,7 +903,7 @@ dependencies = [
 
 [[package]]
 name = "public-api"
-version = "0.52.0"
+version = "0.52.1"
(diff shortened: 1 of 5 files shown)


Enselic (approved):
Thanks, looks good!

Yeah that should be src instead of lib

PR #909, opened 2026-08-04
 Sort trait object bounds (`dyn Foo + Send + Sync`) so that rendering is stable

### LINKED ISSUE #907: Stable/predictable sort order needed for implemented traits
```
--- veilid-core/public-api.txt	2026-07-11 21:13:11
+++ /var/folders/lb/3pt9zbk51rlg_dwsysvzwprr0000gn/T/tmp.eqemavqIZ4	2026-07-16 13:50:31
@@ -1252 +1252 @@
-pub type veilid_core::CryptoSystemGuard<'_>::Target = (dyn veilid_core::CryptoSystem + core::marker::Send + core::marker::Sync)
+pub type veilid_core::CryptoSystemGuard<'_>::Target = (dyn veilid_core::CryptoSystem + core::marker::Sync + core::marker::Send)
```

This module has not changed, but the public_api check reversed the ordering of the `Sync` and `Send` marker traits (it does this randomly), erroneously showing the API as changed when it has not.

Enselic:
Closes #907

### DIFF
--- a/public-api/src/render.rs
+++ b/public-api/src/render.rs
@@ -762,7 +762,7 @@ impl<'c> RenderingContext<'c> {
                 self.render_parenthesized(inputs, output)
             }
             GenericArgs::ReturnTypeNotation => todo!(
-                "can this be triggred in stable rust? if so please report to https://github.com/cargo-public-api/cargo-public-api/issues and include a minimal reproducer"
+                "can this be triggered in stable rust? if so please report to https://github.com/cargo-public-api/cargo-public-api/issues and include a minimal reproducer"
             ),
         }
     }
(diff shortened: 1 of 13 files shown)


public-api/src/render.rs
```diff
@@ -427,6 +427,11 @@ impl<'c> RenderingContext<'c> {
     fn render_dyn_trait(&self, dyn_trait: &rustdoc_types::DynTrait) -> Vec<Token> {
         let mut output = vec![];
 
+        let mut sorted_traits: Vec<_> = dyn_trait.traits.iter().collect();
+        // Sorting on `path` is maybe not ideal, since it isn't always absolute,
```
Enselic, line 431: I don't want to merge this PR before this has been investigated more deeply. Maybe `path` is guaranteed to be absolute nowadays?

If not, I think we need to find a better solution.

As far as I know, the issue this PR fixes is not that bad, since the reordering can only happen when updating nightly Rust version. When changes to types and rendering are expected anyway.

PR #912, opened 2026-09-01
Fix `MINIMUM_NIGHTLY_RUST_VERSION` and add tests enforcing it

wmmc88:
Running `cargo-public-api` v. 0.52.x as a library using the readme example fails because `MINIMUM_NIGHTLY_RUST_VERSION` seems to have fallen out of sync leading to an incompatible unparseable rustdoc json output. This PR updates the minimum nightly rust version to the correct minimum, and adds a test enforcing that the minimum nightly toolchain version from the constant can actually successfully parse the rustdoc json output

### DIFF
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -213,12 +213,7 @@ fn debug_logging() {
 /// * [`MINIMUM_NIGHTLY_RUST_VERSION`] is not set too high
 /// * `cargo pubic-api` suggests the nightly toolchain might be too old when a
 ///   too old nightly toolchain is used
-///
-/// Note: If there is no rustdoc JSON incompatibilities in the previous version,
-/// this test will fail. In that case, feel to add an `#[ignore]` in the
-/// meantime.
 #[test]
-#[ignore = "this test assumes rustdoc JSON incomaptibilities in the previous version but that is not the case right now"]
 fn one_day_before_minimum_nightly_rust_version() {
     test_unusable_toolchain(
         TestCmd::with_proxy_toolchain(&get_toolchain_one_day_before_minimal_toolchain())
--- a/public-api/src/lib.rs
+++ b/public-api/src/lib.rs
@@ -67,7 +67,7 @@ pub use public_item::PublicItem;
 /// The rustdoc JSON format is still changing, so every now and then we update
 /// this library to support the latest format. If you use this version of
 /// nightly or later, you should be fine.
-pub const MINIMUM_NIGHTLY_RUST_VERSION: &str = "nightly-2025-08-02";
+pub const MINIMUM_NIGHTLY_RUST_VERSION: &str = "nightly-2025-11-22";
 // End-marker for scripts/release-helper/src/bin/update-version-info/main.rs
 
 /// See [`Builder`] method docs for what each field means.
--- a/public-api/src/lib.rs
+++ b/public-api/src/lib.rs
@@ -312,3 +312,35 @@ fn deserialize_without_recursion_limit(rustdoc_json_str: &str) -> Result<rustdoc
     deserializer.disable_recursion_limit();
     Ok(serde::de::Deserialize::deserialize(&mut deserializer)?)
 }
+
+#[cfg(test)]
+mod tests {
+    use std::path::Path;
+
+    use super::MINIMUM_NIGHTLY_RUST_VERSION;
+
+    /// The toolchain named by [`MINIMUM_NIGHTLY_RUST_VERSION`] must produce
+    /// rustdoc JSON that this version of the library can parse.
+    #[test]
+    fn minimum_nightly_rust_version_produces_parsable_rustdoc_json() {
+        rustup_toolchain::install(MINIMUM_NIGHTLY_RUST_VERSION).unwrap();
+
+        let build_dir = tempfile::tempdir().unwrap();
+
+        let rustdoc_json = rustdoc_json::Builder::default()
+            .manifest_path(Path::new("../test-apis/comprehensive_api/Cargo.toml"))
+            .toolchain(MINIMUM_NIGHTLY_RUST_VERSION)
+            .target_dir(build_dir.path())
+            .quiet(true)
+            .build()
+            .unwrap();
+
+        if let Err(e) = crate::Builder::from_rustdoc_json(&rustdoc_json).build() {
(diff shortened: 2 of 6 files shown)


wmmc88:
Note: I submitted as two separate commits to show that the example is failing without my fix (ie. in commit [ff3bd4b4584a7f7881132bd3f77c23f1f7596f46](https://github.com/cargo-public-api/cargo-public-api/pull/912/changes/ff3bd4b4584a7f7881132bd3f77c23f1f7596f46)). If the maintainers would like me to squash to two commits together so that every commit individually passes CI, I can do that :) (i just thought it'd be easier to review this way)

cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -213,12 +213,7 @@ fn debug_logging() {
 /// * [`MINIMUM_NIGHTLY_RUST_VERSION`] is not set too high
 /// * `cargo pubic-api` suggests the nightly toolchain might be too old when a
 ///   too old nightly toolchain is used
-///
-/// Note: If there is no rustdoc JSON incompatibilities in the previous version,
-/// this test will fail. In that case, feel to add an `#[ignore]` in the
-/// meantime.
 #[test]
-#[ignore = "this test assumes rustdoc JSON incomaptibilities in the previous version but that is not the case right now"]
 fn one_day_before_minimum_nightly_rust_version() {
```
wmmc88, line 222: The comment justifying the ignore does not make sense to me since if there are "no JSON incompatibilities in the previous version" then the minimum version is not set correctly and is not actually the minimum... The earlier paragraph also explicitly states that the purpose of the test is to make sure the minimum version is not set too high. 

Do the maintainers have some other reason for this or better context?

public-api/src/lib.rs
```diff
@@ -312,3 +312,35 @@ fn deserialize_without_recursion_limit(rustdoc_json_str: &str) -> Result<rustdoc
...
+#[cfg(test)]
+mod tests {
+    use std::path::Path;
+
+    use super::MINIMUM_NIGHTLY_RUST_VERSION;
+
+    /// The toolchain named by [`MINIMUM_NIGHTLY_RUST_VERSION`] must produce
+    /// rustdoc JSON that this version of the library can parse.
+    #[test]
+    fn minimum_nightly_rust_version_produces_parsable_rustdoc_json() {
```
wmmc88, line 325: Without my fix, this test fails: 

```
     Running unittests src\lib.rs (target\debug\build\public-api/92c1dc61d0971853\out\public_api-92c1dc61d0971853.exe)

running 8 tests
test diff::tests::single_and_only_item_added ... ok
test diff::tests::middle_item_added ... ok
test diff::tests::many_identical_items ... ok
test diff::tests::no_off_by_one_diff_skewing ... ok
test diff::tests::middle_item_removed ... ok
test diff::tests::no_diff_means_empty_diff ... ok
test diff::tests::single_and_only_item_removed ... ok
test tests::minimum_nightly_rust_version_produces_parsable_rustdoc_json ... FAILED

failures:

---- tests::minimum_nightly_rust_version_produces_parsable_rustdoc_json stdout ----

thread 'tests::minimum_nightly_rust_version_produces_parsable_rustdoc_json' (105964) panicked at public-api\src\lib.rs:339:13:
rustdoc JSON built with `MINIMUM_NIGHTLY_RUST_VERSION` (`nightly-2025-08-02`) cannot be parsed by this version of `public-api`: missing field `path` at line 1 column 561478


failures:
    tests::minimum_nightly_rust_version_produces_parsable_rustdoc_json

test result: FAILED. 7 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.95s

error: test failed, to rerun pass `-p public-api --lib`
```

cargo-public-api/tests/cargo-public-api-bin-tests.rs
Enselic: To be honest, I don't remember exactly what the problem was. Maybe there were only incompatibilities when using nightly language features, or maybe there were incompatibilities we didn't have tests for and it was a lot of work to write tests.

We can enable the test again and see what happens. I agree with your reasoning.

Can you keep the instructive comment and just comment away the `#[ignore ...` with `// #[ignore ...` though please to make it easy to disable again if needed?

public-api/src/lib.rs
Enselic: Hmm interesting, it seems like a good idea to add a test for this case.

If you don't mind, I might want to have a different form of the test. I'll experiment a bit.

Can you maybe create a new PR just for commenting away the `#[ignore ...` so we can see what happens? Or repurpose this PR for that if you strongly prefer.

Enselic: @wmmc88 If you could take a look at https://github.com/cargo-public-api/cargo-public-api/pull/914, that would be great. Thank you for bringing this to my attention.

The MINIMUM_NIGHTLY_RUST_VERSION_FOR_TESTS mechanism is from 2022, but I think the ergonomics with CI was a lot worse at the time, so it seemed reasonable to do. Now I don't see why we need it, so I removed that mechanism.

With that PR, we now (each night) check that
* MINIMUM_NIGHTLY_RUST_VERSION works
* todays nightly works.

Enselic: I have also confirmed that without bumping MINIMUM_NIGHTLY_RUST_VERSION, that PR fails with the same error as you saw: https://github.com/cargo-public-api/cargo-public-api/pull/914#issuecomment-5498503519

wmmc88: Sounds good! #914 seems to move all the tests to actually testing against the minimum version, so I'm happy to close this PR once that goes in :)

cargo-public-api/tests/cargo-public-api-bin-tests.rs
wmmc88: I just spun up #916 as a test to show what happens when its commented. As expected (and matching my local behavior) it passes since the test is only guarding against the minimum value being set too high (and not that it was set too low.. but that's now covered by #914 ).

public-api/src/lib.rs
Enselic: Turns out today's nightly changes types, but I'll sort that out later today.

Enselic: I got some time already now, so I merged #914.

For the record, I don't think we need to make any new releases, since people can't have relied on the inaccurate value (since then it didn't work).

PR #914, opened 2026-09-01
Bump 0.52.x MINIMUM_NIGHTLY_RUST_VERSION to nightly-2025-11-22

Enselic:
Turns out `nightly-2025-08-02` is not possible to use with 0.52.x. Bump to correct MINIMUM_NIGHTLY_RUST_VERSION.

Also adjust CI so that this mistake can't (read: is unlikely to) happen again. Each night, we now make sure that 
* MINIMUM_NIGHTLY_RUST_VERSION works
* todays nightly works


Enselic:
This PR without bumping MINIMUM_NIGHTLY_RUST_VERSION fails as expected: https://github.com/cargo-public-api/cargo-public-api/pull/915
```
Caused by:
    missing field `path` at line 1 column 646334
```

wmmc88 (approved):

PR #916, opened 2026-09-01
CI TEST(do not merge): remove ignore from test

wmmc88:


Enselic:
Done in https://github.com/cargo-public-api/cargo-public-api/pull/921.

(After reconsidering, I figured it would be good to make a release with the fix to MINIMUM_NIGHTLY_RUST_VERSION, and I plan on doing that this evening.)

PR #918, opened 2026-09-02
ci: Enable macOS in CI again

Enselic:
I'm curious if this works again...

### DIFF
--- a/cargo-public-api/tests/cargo-public-api-bin-tests.rs
+++ b/cargo-public-api/tests/cargo-public-api-bin-tests.rs
@@ -406,9 +406,11 @@ fn minimal_toolchain_works() {
 
 #[test]
 fn warn_when_using_beta() {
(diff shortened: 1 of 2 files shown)


cargo-public-api/tests/cargo-public-api-bin-tests.rs
```diff
@@ -406,7 +406,11 @@ fn minimal_toolchain_works() {
 
 #[test]
 fn warn_when_using_beta() {
-    let mut cmd = TestCmd::with_proxy_toolchain("beta").with_separate_target_dir();
+    rustup_toolchain::install("beta").unwrap();
+    rustup_toolchain::install("nightly").unwrap();
```
Enselic: @copilot both of these are already being installed or already installed. they are no ops. remove them and you will see.

Enselic:
ok, still mysteriously broken...

